PR #22, opened 2014-07-23
Sets the max number of concurrent downloads to 10.

aerych:
Ten might be a little too generous, four or five might be better, but its a starting point that we can tune over time. 
The idea is that we want to prevent apparent network slowness due to the number of concurrent downloads at any given time.


aerych:
This one has gotten too stale and hasn't really been on anyones radar for a while.  Closing.  We can revisit if the need arises.

PR #50, opened 2015-02-03
WPTableViewSectionFooterView fixed width support added

### LINKED ISSUE #49: Table footer view should work with fixed width
`WPTableViewSectionFooterView` should work with fixed width similar to how `WPTableViewSectionHeaderView` does in order to style the footer correctly in iPad and possibly in custom cases.

oguzkocer:
`WPTableViewSectionFooterView` fixed width support added similar to how it works in `WPTableViewSectionHeaderView`. Closes #49.


jleandroperez:
Looks good @oguzkocer!

:shipit:

PR #52, opened 2015-02-05
Fix header and footer width/height calculation

oguzkocer:
There were a few different problems with how we calculated the width and height of the header and footer. I've updated these calculations completely. This is not a perfect solution, although it's a big step up from the previous version. In the future, it would be good to at least remove the duplication of code between footer and header, but as a hotfix this should do for now. With this fix, there should be no overlap between the section header and section content in case the header is multiline.

@aerych & @jleandroperez Can I bother you guys with a quick review?


aerych:
Looks good Oguz!  :shipit:

PR #60, opened 2015-03-02
Issues/47 new colors

### LINKED ISSUE #47: Design: update colors and variables
We finally have a nice color system in WordPress.com for the grays and the naming. I'd like to update the style guide to reflect those changes.

Here are the latest colors: https://wordpress.com/design-handbook/colors/

This will include changing the variable names to be more consistent with WordPress.com. For example, instead of `readGrey`, `whisperGrey`, etc, it would be `grayLight`, `grayDark`, `grayLighten20`, etc.

aerych:
Closes #47


aerych:
cc @sendhil @astralbodies @jleandroperez @bummytime @oguzkocer @diegoreymendez @drw158 
Pinging many since the change affects pretty much the whole app.  When you get a chance, please double check that the updated colors look correct on features you wrangle and sign off below.  I think the easiest way to test would be to just overwrite the version of WPStyleGuide in the pod with the version in this branch. 

- [ ] Notifications
- [ ] Stats
- [ ] Nux
- [ ] Me
- [ ] Reader

Thanks guys!

WordPress-iOS-Shared/Core/WPStyleGuide.h
```diff
@@ -27,48 +27,65 @@
...
 + (UIColor *)wordPressBlue;
-+ (UIColor *)baseLighterBlue;
-+ (UIColor *)baseDarkerBlue;
 + (UIColor *)lightBlue;
-+ (UIColor *)newKidOnTheBlockBlue;
-+ (UIColor *)midnightBlue;
++ (UIColor *)mediumBlue;
++ (UIColor *)darkBlue;
++ (UIColor *)gray;
++ (UIColor *)lightGrey;
```
drw158: Just a note: I don't have a preference between `gray` vs `grey` but let's change them all to the same thing. In the handbook and on the web, we use `gray`, but on Android, everything is grey.

aerych: Oh good catch! Thought I'd kept them consistent.  For fun, http://www.greyorgray.com/

drw158:
Found 2 things in WPStyleGuide.m:

The blue value for `validGreen` should be `102`
https://github.com/wordpress-mobile/WordPress-iOS-Shared/blob/issues/47-new-colors/WordPress-iOS-Shared/Core/WPStyleGuide.m#L246

The red value for `errorRed` should be `217` not `271`
https://github.com/wordpress-mobile/WordPress-iOS-Shared/blob/issues/47-new-colors/WordPress-iOS-Shared/Core/WPStyleGuide.m#L251

drw158:
Also wanted to note, on Android we went ahead and replaced all instances of the deprecated variables and then removed the variables from the file. Not sure what to do here.

aerych:
> we went ahead and replaced all instances of the deprecated variables in the whole project and then removed the variables from the file. Not sure what to do here.

The same, but over a couple of steps so we don't break anything.  Think of this as step one. Then once we update the pod in the app we can clean up the deprecated references. Then finally we'll come back to the pod and delete the old colors here.

aerych:
> Found 2 things in WPStyleGuide.m

Thanks Dave! Updating these.

oguzkocer:
Me page looks good @aerych, thank you!

bummytime:
@aerych Editor looks good.

jleandroperez:
@aerych Notifications look great, thanks Eric!

astralbodies:
@aerych Stats looks good!

aerych:
I double checked reader and nux and they both look fine to me with these changes.

aerych:
@jleandroperez for paranoia sake would you mind taking a look through the code when you have the time?

WordPress-iOS-Shared/Core/WPStyleGuide.m
```diff
@@ -267,16 +277,47 @@ + (UIColor *)keyboardColor {
...
+    return [self darkGrey];
+}
+
+// TODO: Move to fetaure category
++ (UIColor *)nuxFormPlaceholderText {
+    return [self grey];
+}
+
+// TODO: Move to fetaure category
 + (UIColor *)notificationsLightGrey
```
jleandroperez: This one is not currently being used. Can we get it killed? (sorry i missed this before!)

WordPress-iOS-Shared/Core/WPStyleGuide.m
```diff
@@ -267,16 +277,47 @@ + (UIColor *)keyboardColor {
...
 {
-	return [UIColor colorWithRed:244/255.0f green:248/255.0f blue:250/255.0f alpha:1.0f];
+    return [self lightGrey];
 }
 
+// TODO: Move to fetaure category
 + (UIColor *)notificationsDarkGrey
```
jleandroperez: Same as above. Ready to die (SORRY about missing this!)

WordPress-iOS-Shared/Core/WPStyleGuide.m
```diff
@@ -374,4 +419,67 @@ + (void)configureFollowButton:(UIButton *)followButton {
     [followButton setTitleColor:[self allTAllShadeGrey] forState:UIControlStateNormal];
 }
 
+
+#pragma mark - Deprecated Colors
+
++ (UIColor *)baseLighterBlue __deprecated
```
jleandroperez: This is *SO* cool!

jleandroperez:
@aerych two minor comments. There are are two Notification colors that aren't being used anywhere in the app.

Other than that, this is flawless!. :shipit:

aerych:
Thanks @jleandroperez!  I'll remove those two colors real quick.

WordPress-iOS-Shared/Core/WPStyleGuide.m
```diff
@@ -139,111 +140,120 @@ + (UIFont *)tableviewSectionHeaderFont
     return [WPFontManager openSansBoldFontOfSize:12.0];
 }
 
+
 #pragma mark - Colors
+// https://wordpress.com/design-handbook/colors/
 
-+ (UIColor *)wordPressBlue
++ (UIColor *)colorWithR:(NSInteger)red G:(NSInteger)green B:(NSInteger)blue alpha:(CGFloat)alpha
```
sendhil: Why did we create this method? It seems like it saves a few characters but I think it makes readability worse in the file.

aerych: > but I think it makes readability worse in the file.
Really? For me its so much easier to understand RGB values when I can just focus on the number, and not have to filter out all the repetitious division.

sendhil: I think most objective c developers are used to the division by now as it's pretty common to see the `/255.0` littered throughout Objective-C. Why not make the naming something a little more natural and closer to the original method so that way it reads more natural and muscle memory kicks in. For example:

`[UIColor colorWithRedValue:colorWithGreenValue:colorWithBlueValue:alpha:]`

sendhil: I should add, when i see the `colorWithR:` I still pause for a second to think and translate R in my head to Red whereas when I see `colorWithRed:` or `colorWithRedValue:` i don't really stop to think about it as it reads more naturally and doesn't need any translation.

aerych:
> I think most objective c developers are used to the division by now 

Maybe, but you're making an assumption. Even if people are used to it that's no reason we shouldn't try to make improvements :) 

>Why not make the naming something a little more natural and closer to the original method so that way it reads more natural and muscle memory kicks in.

The goal here was to type less. :)

> I should add, when i see the colorWithR: I still pause for a second to think and translate R in my head to Red whereas when I see colorWithRed: or colorWithRedValue: i don't really stop to think about it as it reads more naturally and doesn't need any translation.

Its more natural because that's what you're used to, of course. :)

Maybe its just a matter of individual preference.  I think of colors as RGB values, so "R" "G" "B" as args make sense to me.

WordPress-iOS-Shared/Core/WPStyleGuide.m
aerych: > I think most objective c developers are used to the division by now 

Maybe, but you're making an assumption. Even if people are used to it that's no reason we shouldn't try to make improvements :) 

>Why not make the naming something a little more natural and closer to the original method so that way it reads more natural and muscle memory kicks in.

The goal here was to type less. :)

> I should add, when i see the colorWithR: I still pause for a second to think and translate R in my head to Red whereas when I see colorWithRed: or colorWithRedValue: i don't really stop to think about it as it reads more naturally and doesn't need any translation.

Its more natural because that's what you're used to, of course. :)

Maybe its just a matter of individual preference.  I think of colors as RGB values, so "R" "G" "B" as args make sense to me.

sendhil:
>Maybe, but you're making an assumption. Even if people are used to it that's no reason we shouldn't try to make improvements :)

This is fair.

>The goal here was to type less. :)

I would rather type more if it makes my code more readable.

>Its more natural because that's what you're used to, of course. :)

Exactly - I'm someone coming in reading your code and it forces me to pause and think unnecessarily. One of our top priorities with our code is to make it easy to read for our fellow developers and if more people are used to the original syntax than they are the new syntax you are going to have more people who as they scan this code have to pause and translate than you will who scan and naturally have it click.

>Maybe its just a matter of individual preference. I think of colors as RGB values, so "R" "G" "B" as args make sense to me.

RGB in and of itself makes sense but the way the method is laid out the characters RGB aren't right next to each other, they are split across parameter names which for me(and I'd imagine others) means you have to pause and think about it for a moment.

aerych:
> I would rather type more if it makes my code more readable

I know you're trying to sell me that the method name isn't very readable... its a hard sell coming from one voice and could a matter of taste. I'd be happy to hear other opinions and if it makes sense to tweak the method definition then I'll happily give the PR a :shipit:  :) 

> One of our top priorities with our code is to make it easy to read for our fellow developers

Getting rid of repetitive division does make the code easier to read :) 

> if more people are used to the original syntax than they are the new syntax you are going to have more people who as they scan this code have to pause and translate than you will who scan and naturally have it click

This is a common push back to change: "its different so I have to think about it". If you think about it once or twice and then adapt then if the change helps you be more efficient then maybe its a good change.  If you keep having to think about it over and over, then probably it wasn't a good change.  But we can't peer into the future to know. :P

sendhil:
>I know you're trying to sell me that the method name isn't very readable... its a hard sell coming from one voice and could be a matter of taste.

It's the voice of someone who came in and read your code. You changed a fairly standard thing in objective c(`[UIColor colorWithRed:...`]) to something non standard based on a matter of taste. If this was just a project for one set of eyes I think that'd be fine but when you have as many eyes on the project as ours does I don't think personal taste is a good enough reason to change it if that change will introduce introduce friction as people read the code.

>Getting rid of repetitive division does make the code easier to read :)

This is true especially as someone could accidentally introduce a bug if they forgot a `.0` at the end or didn't type `255` and put something else in. That being said  introducing a non standard naming scheme does not make it easier to read so why not do the best of both worlds and adjust the method signature to something more standard  as you aren't going to find many objective c method signatures following this pattern.

>This is a common push back to change: "its different so I have to think about it".

While this is sometimes true, under that rationale we should just accept change because it's change without evaluating whether it's a good one or not. I like change when it's helpful and beneficial - I agree with you that cutting out the manual division is a good thing, but I don't agree about adopting a non standard signature.

>If you think about it once or twice and then adapt then if the change helps you be more efficient then maybe its a good change.

This would be true if every other codebase in objective c adopted the same change but the reality is we are probably going to see a lot more of the more verbose syntax than we will the shortened one forcing us to jump back and forth between two different naming schemes creating friction at times when we read the code which is not ideal unless there was a compelling reason to do so which I don't see in this case.

diegoreymendez:
Hey guys.

It sounds like a good thing not to have to use float values for colors.  OTOH I agree that we should try to avoid abbreviating on method names.

Maybe there's a middle ground we can use such as:

`colorWithIntegerRed:green:blue:alpha`

Or any other option you can think of @aerych that more clearly explains the difference with the standard `colorWithRed:green:blue:alpha`.

Since we have auto-completion, the longer name would not be a big issue.

aerych:
> You changed a fairly standard thing in objective c([UIColor colorWithRed:...]) to something non standard based on a matter of taste

And making the code easier to read, and shorter to type. :)

> a non standard naming scheme does not make it easier to read

I don't buy the assertion I've used a "non-standard naming scheme".  What does that even mean? 

> why not do the best of both worlds and adjust the method signature

I think I already said I'm fine with that :) 

> but the reality is we are probably going to see a lot more of the more verbose syntax than we will the shortened one forcing us to jump back and forth between two different naming schemes creating friction at times when we read the code 

Not sold that there is any friction whatsoever as the naming made perfect sense to me at the time, and was not commented on during the review of the PR. :)

aerych:
Hey @diegoreymendez thanks for chiming in :)

> Maybe there's a middle ground we can use such as: `colorWithIntegerRed:green:blue:alpha`

Sounds fine to me.

PR #61, opened 2015-03-04
Update bundle usage for cocoapods 0.36

sendhil:
As per:
cocoapods-updates

cc @diegoreymendez

### DIFF
--- a/WordPress-iOS-Shared/Core/WPFontManager.m
+++ b/WordPress-iOS-Shared/Core/WPFontManager.m
@@ -172,7 +172,7 @@ static NSString * const kBundle = @"WordPress-iOS-Shared.bundle";
 + (void)dynamicallyLoadFontResourceNamed:(NSString *)name
 {
     NSString *resourceName = [NSString stringWithFormat:@"%@/%@", kBundle, name];
-    NSURL *url = [[NSBundle bundleForClass:[WPFontManager class]] URLForResource:resourceName withExtension:@"ttf"];
+    NSURL *url = [[NSBundle bundleForClass:self] URLForResource:resourceName withExtension:@"ttf"];
     NSData *fontData = [NSData dataWithContentsOfURL:url];
     
     if (fontData) {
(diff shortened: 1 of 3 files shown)


WordPress-iOS-Shared/Core/WPFontManager.m
```diff
@@ -172,7 +172,7 @@ + (UIFont *)merriweatherBoldFontOfSize:(CGFloat)size
 + (void)dynamicallyLoadFontResourceNamed:(NSString *)name
 {
     NSString *resourceName = [NSString stringWithFormat:@"%@/%@", kBundle, name];
-    NSURL *url = [[NSBundle mainBundle] URLForResource:resourceName withExtension:@"ttf"];
+    NSURL *url = [[NSBundle bundleForClass:[WPFontManager class]] URLForResource:resourceName withExtension:@"ttf"];
```
diegoreymendez: Since you're in a static method in WPFontManager you might as well replace `[WPFontManager class]` with `self`.  This is not really a mistake though.

diegoreymendez: PS: in static methods `self` refers to the class, and `self.class == self`.  So what I'm saying is that `self` is enough, and `self.class` is sort of redundant.

diegoreymendez:
Just a comment above, which you can opt to ignore, if you prefer.

We should be careful before upgrading to Cocoapods 0.36 since AFNetworking seems to rely on `mainBundle` too.

WordPress-iOS-Shared/Core/WPFontManager.m
sendhil: Fixed in https://github.com/wordpress-mobile/WordPress-Shared-iOS/commit/c17b0bd12e2e27ea7d4338c734a3eea0d0a94f6e

sendhil:
@diegoreymendez  - yea I spotted that with AFNetworking, bleh. We'll definitely wait on upgrading to 0.36 until they are compatible.

PR #69, opened 2015-07-08
Fixes WPTableViewSectionFooterView height calculation

jleandroperez:
We were using a different font for the Height calculation. In this PR we're simply normalizing both fonts, and updating the Podspec as well.

Needs Review: @diegoreymendez (Thank you Diego!)


diegoreymendez:
:shipit:

jleandroperez:
Thank you sir!

PR #75, opened 2015-07-30
Unify WPTableViewSectionHeader and WPTableViewSectionFooterView

jleandroperez:
Since both classes contain (roughly) the exact same code... we're simplifying things, right here.

**WPTableViewSectionHeaderFooterView** is a **UITableViewSectionHeaderFooterView** subclass, with few extra properties and helpers.

By default, it implements two modes: Header / Footer (which will effectively set the appropiate style to be displayed as a Header / Footer).

Users may also specify custom properties (Font / Insets), so that we don't need to patch the Shared Repository everytime a screen requires a custom tableView Section Header (or Footer).

Needs Review: @SergioEstevao (Thanks in advance Sergio!)


WordPress-iOS-Shared/Core/WPTableViewSectionHeaderFooterView.m
```diff
@@ -0,0 +1,227 @@
...
+@end
+
+
+@implementation WPTableViewSectionHeaderFooterView
+
+- (instancetype)initWithStyle:(WPTableViewSectionStyle)style
+{
+    NSParameterAssert(style == WPTableViewSectionStyleHeader || style == WPTableViewSectionStyleFooter);
+    
+    self = [super init];
```
SergioEstevao: You should invoke the designated initializer of your base class here: initWithReuseIdentifier. If you aren't make the reuse identifier available for this class you can invoke with nil.

On the same page you should override that default initializer to use this initializer with a default style set.

WordPress-iOS-Shared/Core/WPTableViewSectionHeaderFooterView.m
```diff
@@ -0,0 +1,227 @@
...
+        titleLabel.backgroundColor      = [UIColor clearColor];
+        titleLabel.shadowOffset         = CGSizeZero;
+        [self addSubview:titleLabel];
+
+        // Background
+        UIView *backgroundView          = [UIView new];
+        backgroundView.backgroundColor  = [UIColor clearColor];
+        
+        // Initialize Prperties
+        self.uppercase                  = isHeader;
```
SergioEstevao: You should use the ivars instead of the properties methods on the initializer, but I assume you want the side effects of the properties changes to apply?

WordPress-iOS-Shared/Core/WPTableViewSectionHeaderFooterView.m
```diff
@@ -0,0 +1,227 @@
...
+#pragma mark - Overriden Properties
+
+/**
+    Note:
+    The purpose of the following overrides are to prevent UI glitches, caused by users accessing the 
+    default textLabel and detailTextLabel provided by the superclass.
+    We're effectively disabling those fields!
+ */
+
+- (UILabel *)textLabel
```
SergioEstevao: Really like this!

WordPress-iOS-Shared/Core/WPTableViewSectionHeaderFooterView.m
```diff
@@ -0,0 +1,227 @@
...
+- (void)layoutSubviews
+{
+    [super layoutSubviews];
+
+    CGFloat sectionWidth    = CGRectGetWidth(self.bounds);
+    CGFloat titleWidth      = [[self class] titleLabelWidthFromSectionWidth:sectionWidth titleInsets:self.titleInsets];
+    CGSize titleSize        = [[self class] sizeForTitle:self.titleLabel.text titleWidth:titleWidth font:self.titleFont];
+    CGFloat padding         = (sectionWidth - titleWidth) * 0.5;
+
+    self.titleLabel.frame   = CGRectIntegral(CGRectMake(padding, _titleInsets.top, titleWidth, titleSize.height));
```
SergioEstevao: why use the _titleInsets ivar here?

WordPress-iOS-Shared/Core/WPTableViewSectionHeaderFooterView.m
jleandroperez: Exactly, in this particular case, i'm afraid it's on purpose. Hitting the ivars directly wouldn't have the same effect, i'll add a comment.

Thank you!

WordPress-iOS-Shared/Core/WPTableViewSectionHeaderFooterView.m
jleandroperez: Thank you sir!

WordPress-iOS-Shared/Core/WPTableViewSectionHeaderFooterView.m
jleandroperez: You're right, fixing! Thank you!!

WordPress-iOS-Shared/Core/WPTableViewSectionHeaderFooterView.m
jleandroperez: Ahhhhhh awesome catch, i absolutely missed that. Updating accordingly, thanks sergio!

jleandroperez:
@SergioEstevao ready for a second round. Thank you!

WordPress-iOS-Shared/Core/WPTableViewSectionHeaderFooterView.m
```diff
@@ -0,0 +1,243 @@
...
+@end
+
+
+@implementation WPTableViewSectionHeaderFooterView
+
+/*
+You should invoke the designated initializer of your base class here: initWithReuseIdentifier.
+If you aren't make the reuse identifier available for this class you can invoke with nil.
+
+On the same page you should override that default initializer to use this initializer with a default style set.
```
SergioEstevao: You should add this:
```
- (instancetype)initWithReuseIdentifier:(NSString *)reuseIdentifier
{
return [self initWithReuseIndentifier:reuseIdentifier style:WPTableViewSectionStyleHeader]
}
```
to avoid people calling the super initialiser and not getting the state variable properly set and the subviews set.

jleandroperez:
@SergioEstevao `implementsWithReuseIdentifier` added, thank you! (Github killed the comment after pushing the commit, sorry about the extra noise)

SergioEstevao:
:shipit:

jleandroperez:
Thank you sir!

PR #77, opened 2015-08-04
Issues/76 update merriweather

aerych:
Updates the version of Merriweather.  Easiest way to identify the change is by observing the descender in a capital Q. The correct font will show a smaller descender.  The OTF fonts previously used were the correct version, but they are removed in order to use TTFs for all. 

#### Previous:
merriweather-previous

#### Now: 
merriweather-new-1
merriweather-new-2

Needs Review: @diegoreymendez


aerych:
Added the DO NOT MERGE label.  @mnt has discovered that this version of Merriweather from Google, which is used on wpcom, is actually the older version and the TTF currently in the shared pod are more recent (but not the OTFs).  
We'll either keep this PR for consistency or we'll drop it and update the OTFs instead, depending on what is decided for wpcom.

mnt:
After looking into this a bit more, I've decided to stick with the Google version of the font. The italics in the  new version have not been updated to match the heavier weight or redrawn letterforms of the roman styles (you can see it in the first screenshot above: bold italic is lighter than bold; the Q doesn't match between the two, either).

We'll want to proceed with this change, but take note of where we're currently using Merriweather Light, as we'll likely want to switch it to Regular because of the difference in weight between the two light versions.

aerych:
Ready for review @diegoreymendez

diegoreymendez:
Everything I checked is in place.  The Q note was helpful to check the font assets manually.  Thanks, @aerych.

:shipit:

aerych:
Thanks @diegoreymendez!

PR #81, opened 2015-11-10
Adds Cocoapods framework support

koke:
Two main changes here:

1. Sets `header_dir` to WordPressShared to avoid problems with
underscores/dashes in names. I bumped the version to 0.5 because this
changes how you import headers from this pod.

2. Made WPAnimatedImageResponseSerializer non-public. Since it's header
file imports from AFNetworking (because it's a subclass), it was giving
errors about importing a non-modular header. I think the proper fix
would have been to switch to `@import` but that would only work if
frameworks are enabled. In any case, WPAnimatedImageResponseSerializer
isn't used outside WPImageSource, so we can revisit this if we need to
reuse that class somewhere else.

See https://github.com/wordpress-mobile/WordPress-iOS/issues/4434

As I bumped version, it'll need to be released after merging

Needs Review: @jleandroperez


jleandroperez:
Looks good sir!

:shipit:

PR #85, opened 2015-12-18
Deprecate WPTableViewSectionHeaderFooterView

koke:
WPTableViewSectionHeaderFooterView started its life as a way to customize the section headers to adapt to iOS 7 design changes, and has grown into a different beast today.

Of all the things it does, we just care about using a custom font and color, which can be achieved without custom classes, using the delegate method `tableView:willDisplayHeaderView:forSection:`.

I've marked WPTableViewSectionHeaderFooterView as deprecated and it should be removed in a next version once everything is migrated.

I've also added a couple new methods to call on `tableView:willDisplayHeaderView:forSection:` and `tableView:willDisplayFooterView:forSection:`.

For example usage see https://github.com/wordpress-mobile/WordPress-iOS/pull/4591


SergioEstevao:
:shipit:

PR #87, opened 2016-02-26
No more open sans

koke:
Adds new helper methods for system font, and deprecates usage of Open Sans

Question: instead of having the font name (or system) in the method names, should it be more generic? I'm thinking `interface*FontOfSize` and `content*FontOfSize`, so nothing but `WPFontManager` has to care if we ever switch fonts again.

See: https://github.com/wordpress-mobile/WordPress-iOS/issues/4895
Cc: @mattmiklic 
Needs Review: @astralbodies


mattmiklic:
@koke I've taken a look and so far everything looks good visually with the font change. There are a few places I'd like to make some small adjustments, but they are not urgent so I'll make a separate issue once I've had a chance to give it a more thorough look.

astralbodies:
:shipit:

PR #90, opened 2016-04-11
Replace AFNetworking dependency for NSURLSession

SergioEstevao:
This PR replaces the use of AFNetworking for NSURLSession on the WPImageSource class.

It also updates the current tests to use XCTestExpectations and to use the latests version OHHTTPStubs in order to support NSURLSession methos.

How to test:
 * Run the unit tests and make sure they are ok.
 * Integrate on the main app and check stats, gravatar, the WordPress media library, and the reader.
 
Needs Review: @astralbodies


astralbodies:
@SergioEstevao Lets also get another team member in on this review to make sure it gets enough eyes :smile:

SergioEstevao:
I also think @jleandroperez maybe interested :).

SergioEstevao:
@astralbodies @aerych @jleandroperez, I create this branch: **test_wpshared_with_nsurlsession**  on the main WP iOS app to make it easier to test it.

aerych:
Changes look good to me. I took the test branch for a spin (thanks for that!!) and viewed a few image heavy posts in the reader.  Everything loaded fine.  Comments in the reader also loaded fine.  Animated gifs are showing a single still frame as expected.  
Very nice @SergioEstevao :) 
:shipit: from me.

WordPress-iOS-Shared/Core/WPImageSource.m
```diff
@@ -24,17 +25,19 @@ + (instancetype)sharedSource
 
 - (void)dealloc
 {
-    [_downloadingQueue cancelAllOperations];
+    [_downloadsSession invalidateAndCancel];
 }
 
 - (id)init
```
jleandroperez: I know this wasn't added here.. but can we have `instancetype` here perhaps?

WordPress-iOS-Shared/Core/WPImageSource.m
```diff
@@ -24,17 +25,19 @@ + (instancetype)sharedSource
...
 {
     self = [super init];
     if (self) {
-        _downloadingQueue = [[NSOperationQueue alloc] init];
         _urlDownloadsInProgress = [[NSMutableSet alloc] init];
         _successBlocks = [[NSMutableDictionary alloc] init];
         _failureBlocks = [[NSMutableDictionary alloc] init];
+
+        NSURLSessionConfiguration *configuration = [NSURLSessionConfiguration defaultSessionConfiguration];
+        _downloadsSession = [NSURLSession sessionWithConfiguration:configuration];
```
jleandroperez: Just wondering if we don't wanna use the backgroundSessionConfiguration, which would allow us to finish downloading the assets, even when the app dies. Or not! just a thought

WordPress-iOS-Shared/Core/WPImageSource.m
```diff
@@ -82,33 +85,29 @@ - (void)startDownloadForURL:(NSURL *)url authToken:(NSString *)authToken {
...
-				return;
-			}
-			[self downloadedImage:image forURL:url];
-		} failure:^(AFHTTPRequestOperation *operation, NSError *error)
-		{
-			[self downloadFailedWithError:error forURL:url];
-		}];
-		
-        [_downloadingQueue addOperation:operation];
+        NSURLSessionDownloadTask *task = [self.downloadsSession downloadTaskWithRequest:request
```
jleandroperez: *BEAUTIFUL*. THANK YOU for so much awesomeness @SergioEstevao

jleandroperez:
@SergioEstevao two minor comments added. Awesome touch, thanks for nuking AF!!

:shipit: on my side sir!

WordPress-iOS-Shared/Core/WPImageSource.m
koke: Looking at the current use cases, I'd say we don't want background downloads as the images being downloaded aren't important enough. If that becomes useful, the caller should be able to inject the session/session configuration, but that's something to revisit when we use NSURLSession across the app.

SergioEstevao: Thank you for the answering for me @koke I agree that background downloads are not desirable here. And even if we wanted them will not be easy to implement because then you cannot use the session block interface, and you need to complete callbacks for background downloads if the the app is not running.

PR #91, opened 2016-04-14
Add WordPressShared framework target

koke:
So far the project was building a static library. There are plenty of reasons to switch to frameworks but what triggered this was trying to add Swift code and Swift unit tests.

This PR:

- Adds a new `WordPressShared` framework target/scheme
- Bumps the iOS deployment target to 8.0, since it's the minimum that supports frameworks
- Switch the unit tests to use the framework instead of the static library
- Adds a small unit test in Swift to verify that it's working
- Removed `IS_IPAD` and other macros from the PCH file and converted them to regular methods in `WPDeviceIdentification`
- Switch travis to build the new scheme with xcodebuild, as that's what we're using everywhere else.

Needs Review: @SergioEstevao

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-osx_image: xcode7.1
+osx_image: xcode7.3
 language: objective-c
 script:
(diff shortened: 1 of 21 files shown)


SergioEstevao:
:shipit:

PR #93, opened 2016-04-18
WPImageSource refactor with OperationPool

koke:
This is the first of a series of refactors aimed at reducing duplication in our image downloading code. I've started with WPImageSource.

The main functionality of WPImageSource was to prevent duplicate requests. This was originally created with lists of comments in mind, where it was easy to see multiple instances of the same avatar. Requesting the same avatar twice in a row would generate two identical requests, as caching only takes effect after the first successful download.

But this behavior isn't really specific to images, and could be effectively reused to prevent duplicate requests of any other type. So I've extracted that logic into `OperationPool`. 

Besides the request coalescing, WPImageSource also handled request authentication and some security, which I've moved to some NSURL extensions, as I've also seen them duplicated recently.

Lastly, I've introduced a `Result` type. It's very lightweight and really simplifies moving transforming and moving data from one process to another where errors might occur along the way. 

Needs Review: @SergioEstevao


WordPress-iOS-Shared/Core/OperationPools.swift
```diff
@@ -0,0 +1,286 @@
...
+///
+/// When an operation finishes, all the completion blocks are queued
+/// asynchronously at the same time.
+///
+/// Note that there's a delay between the time the `start` function calls its
+/// `completion` function and the completion blocks are called. If you want to
+/// add some caching of results, the recommended way is to handle storing the
+/// cached values from the `start` function to avoid duplicate operations.
+///
+public class OperationPool<Key: Hashable, Value, Error: ErrorType> {
```
diegoreymendez: Careful there.  Hashes are NOT unique, nor required to be.

WordPress-iOS-Shared/Core/OperationPools.swift
```diff
@@ -0,0 +1,286 @@
...
+                return .failure(error as NSError)
+            }
+        }
+        super.init(parser: parser, session: session)
+    }
+}
+
+/// Errors related to OperationPool.
+/// 
+public enum OperationPoolError: Int, ErrorType {
```
diegoreymendez: Does this need to be public? (ie accessible from outside the pod?)

WordPress-iOS-Shared/Core/OperationPools.swift
```diff
@@ -0,0 +1,286 @@
...
+    /// that UIImage(data:) could process.
+    ///
+    case invalidImageData
+
+    /// A description of the error, in English.
+    ///
+    /// - important: This description is for debugging purposes only and should
+    /// not be shown to the user.
+    ///
+    public var description: String {
```
diegoreymendez: Why var?

WordPress-iOS-Shared/Core/OperationPools.swift
```diff
@@ -0,0 +1,286 @@
...
+        case .noDataAndNoError:
+            return "Network response had no data and no error"
+        case .invalidImageData:
+            return "Couldn't generate a valid image with the returned data"
+        }
+    }
+
+    /// The error converted to a NSError.
+    ///
+    public var nserror: NSError {
```
diegoreymendez: Same as above regarding `public`.

WordPress-iOS-Shared/Core/OperationPools.swift
diegoreymendez: In the case of `NSURLRequest` objects for example: the full URL would be unique, whereas the hash would not - meaning using the full URL as the key would be preferred to using the hash.

koke: You almost scared me for a second 😉 
Yes, it's possible that two keys would have the same hash (although unlikely). It's also not a problem.

The `Key` is `Hashable` because we're using a dictionary internally, and it's a requirement. `Hashable` also requires `Equatable`, so we can have a way to check if those objects that have the same value are actually equal. `Hashable` is an optimization so a `Dictionary` can do fast lookup, but it still verifies equality.

Just to double check I ran this on a playground:

```swift
struct FakeHash: Hashable {
    let value: Int
    var hashValue: Int {
        return 1
    }
}

func == (lhs: FakeHash, rhs: FakeHash) -> Bool {
    return lhs.value == rhs.value
}

let item1 = FakeHash(value: 1)
let item2 = FakeHash(value: 2)
var dict = [FakeHash: String]()
dict[item1] = "1"
dict[item2] = "2"

item1 == item2
// false
item1.hashValue == item2.hashValue
// true
dict.description
// [FakeHash(value: 1): "1", FakeHash(value: 2): "2"]
```

WordPress-iOS-Shared/Core/OperationPools.swift
koke: If you want to do any error checking, yes. Otherwise, you'd get an opaque `ErrorType` with no way to check what the error was. Consider this the Swift equivalent of Obj-C error code enums and `extern NSString * const OperationPoolErrorDomain` in header files.

WordPress-iOS-Shared/Core/OperationPools.swift
koke: It's a computed property.

WordPress-iOS-Shared/Core/OperationPools.swift
koke: I think it's useful that this is `public`. All the existing OperationPool implementations use `NSError` as their error type. This means that any requester can get a Result with a NSError in it.

You want to check, for instance, if the error is `.invalidImageData`. You then have a `OperationPoolError` and a `NSError` with no way to compare them.

WordPress-iOS-Shared/Core/OperationPools.swift
diegoreymendez: We don't return the `ErrorType`, though.  But, yes this can be used to compare the NSError's code.

However... since we're using `ErrorType` why not just stick to Apple's new default of using exceptions for error handling?  It's weird we have an `ErrorType` object but instead return an `NSError`.

koke: I think I understand your point as: this conforms to `ErrorType` but it's being used as `NSError`.
And now I realize that `OperationPoolError` doesn't really need to conform to `ErrorType`, and it might be slightly less confusing if I remove that.

I believe the alternative to this would be to define a couple of separate `NSError` constants, but it feels slightly better to group errors inside an `enum`.

`ErrorType` is a bit weird in general.

We're kind of stuck with `NSError` since that's what we get from `NSURLSession`. An alternative would be to wrap those returned errors inside a `OperationPoolError` value, but I don't think it makes much sense: if there's an error it's probably going to come from `NSURLSession`, our custom errors are only there as sanity checks. So we don't want to be wrapping and unwrapping the most common error cases.

> why not just stick to Apple's new default of using exceptions for error handling? 

Because you can't `throw` an error asynchronously.

koke: I removed the `ErrorType` last night but then reverted it. In Swift, `ErrorType` is an empty protocol, which just means that a type represents some kind of error.

Swift automatically bridges ErrorType to NSError, but it's not publicly exposed so we can't override how it does it.

screen shot 2016-04-20 at 08 39 45

I think a good idea would be to expose `domain`/`code`. If we wanted to check for a specific error, with the current version:
```swift
pool.request(request) { result in
    if let error = result.error
        where error == OperationPoolError.noDataAndNoError.nserror {
        // handle specific error
    }
}
```

Exposing domain/code:

```swift
pool.request(request) { result in
    if let error = result.error
        where error.domain == OperationPoolError.domain
            && error.code == OperationPoolError.noDataAndNoError.code {
        // handle specific error
    }
}
```

I think I prefer the first version though, as it's more direct in its comparison.

An even more _classic_ approach would be to define the error enum like this:

```swift
public enum OperationPoolError {
    public static let domain = "org.wordpress.shared.operationpool"
    public enum Code: Int {
        case noDataAndNoError
        case invalidImageData
    }

    private static var noDataAndNoError: NSError {
        return NSError(domain: domain, code: Code.noDataAndNoError.rawValue, userInfo: [NSLocalizedDescriptionKey: "Network response had no data and no error"])
    }
    private static var invalidImageData: NSError {
        return NSError(domain: domain, code: Code.invalidImageData.rawValue, userInfo: [NSLocalizedDescriptionKey: "Couldn't generate a valid image with the returned data"])
    }
}
```

And this is how you'd check:

```swift
pool.request(request) { result in
    if let error = result.error
        where error.domain == OperationPoolError.domain
            && error.code == OperationPoolError.Code.noDataAndNoError.rawValue {
        // handle specific error
    }
}
```

But checking for an error gets even more verbose.

koke:
I'm having second thoughts about the `OperationPool` name. I think the best fit for the second part is `Proxy`. As for the first, `Operation` would be good if it weren't to be confused with `NSOperation`.

I'm also reconsidering if we should drop `completionQueue` and leave that responsibility to the specific implementations.

SergioEstevao:
Is this PR still active @koke ? or should we close it?

koke:
As much as it pains me, I'm going to close it for now. I'm not going to be working on this or refactoring image loading for a while. At least this is new code, so we should be able to reuse it without conflicts if we reopen it in the future.

PR #94, opened 2016-05-04
Improved the code to properly hide the accessoryView in WPNoResultsView.

diegoreymendez:
### Description:

Fixes [this issue](https://github.com/wordpress-mobile/WordPress-iOS/issues/5293), which is caused by the code not enforcing some GUI layout rules in the appropriate spots.

If the fix is accepted a follow-up PR will boost the pod version and properly include it in WPiOS.

### Testing:

1. Checkout WPiOS branch `issue/5293-no-results-accessory-view-problems`.
2. Reinstall the pods
3. Launch the app.
4. Go to the list of posts or pages for any of your blogs.
5. Make sure you select a filter with no results.
6. In portrait mode, the ink well image should be visible at all times.
7. In landscape mode, the ink well image should not be visible ever.  Do a pull-to-refresh on the list and make sure it stays that way.

Please review @aerych 

/cc @astralbodies

### DIFF
--- a/WordPress-iOS-Shared-Example/Podfile.lock
+++ b/WordPress-iOS-Shared-Example/Podfile.lock
@@ -7,7 +7,7 @@ PODS:
     - CocoaLumberjack/Core
   - CocoaLumberjack/Extensions (2.2.0):
     - CocoaLumberjack/Default
-  - WordPress-iOS-Shared (0.5.5):
+  - WordPress-iOS-Shared (0.5.6):
     - CocoaLumberjack (~> 2.2.0)
 
 DEPENDENCIES:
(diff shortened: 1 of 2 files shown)


aerych:
Hi @diegoreymendez 
It looks like the layout on the iPad will be off center after this change.  You can reproduce it by rotating the ipad while the no results view is displayed. 
simulator screen shot may 4 2016 5 41 31 pm

diegoreymendez:
Thanks for the feedback Eric.  I'll be working on it.

diegoreymendez:
@aerych - I restored most of the rotation code, which I believe should fix the rotation issue.  This is ready for another look.

aerych:
One of these days i want to rebuild this thing with stack views and kill all the frame setting in `layoutSubviews`.   But, something for another day.

aerych:
:shipit: yo :)

diegoreymendez:
Thanks Mr. Eric!

PR #97, opened 2016-05-11
UIColor hexString now works with grayscale colors.

frosty:
I've updated `hexString` to support colors with only two components. This is needed for a new helper I'm adding to WPiOS which allows easier styling of attributed strings via HTML (requiring CSS / hex colors). Without this fix, `hexString` won't work with e.g. `UIColor.whiteColor()`.

To test: 

* Verify that the `UIColorHelpersTests` pass (and make sense!).
* Check the code :)

Needs review: @diegoreymendez  
Thanks!

### DIFF
--- a/WordPress-iOS-Shared/Core/UIColor+Helpers.m
+++ b/WordPress-iOS-Shared/Core/UIColor+Helpers.m
@@ -50,24 +50,34 @@
 - (NSString *)hexString
 {
     // Taken from http://stackoverflow.com/questions/8689424/uicolor-to-hexadecimal-web-color
-    
+
     NSString *webColor = nil;
-    
-    // This method only works for RGB colors
-    if (CGColorGetNumberOfComponents(self.CGColor) == 4)
+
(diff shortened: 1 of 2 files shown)


frosty:
@diegoreymendez: Jorge has some issues with his CocoaPods installation at the moment. Would you mind picking this up? Thanks!

WordPress-iOS-Shared/Core/UIColor+Helpers.m
```diff
@@ -50,24 +50,34 @@ + (UIColor *)colorWithHexString:(NSString *)hex {
...
-    
+
     NSString *webColor = nil;
-    
-    // This method only works for RGB colors
-    if (CGColorGetNumberOfComponents(self.CGColor) == 4)
+
+    size_t componentCount = CGColorGetNumberOfComponents(self.CGColor);
+
+    // RGB colors
```
diegoreymendez: Should probably be RGBA for clarity.

diegoreymendez:
:shipit:

PR #106, opened 2016-08-02
WPNoResultsView now takes into account table header and footer views

frosty:
Fixes https://github.com/wordpress-mobile/WordPress-iOS/issues/5765 by adjusting the vertical position of `WPNoResultsView` to account for table header and footer views.

To test:

* Using this branch, ensure that the no results view appears vertically centered in all locations that call `centerInSuperview` in WPiOS.
* Ensure that the no results view isn't partially hidden on the posts list on an iPhone 4s.

Needs review: @aerych

### DIFF
--- a/WordPress-iOS-Shared/Core/WPNoResultsView.m
+++ b/WordPress-iOS-Shared/Core/WPNoResultsView.m
@@ -203,29 +203,40 @@
 }
 
 - (void)centerInSuperview {
-    
+
     if (!self.superview) {
         return;
     }
-    
+
     // Center in superview
     CGRect frame = [self superview].frame;
-    
+
     // account for content insets of superview if it is a scrollview
     if ([self.superview.class isSubclassOfClass:[UIScrollView class]]) {
         UIScrollView *scrollView = (UIScrollView *)self.superview;
-        
+
         CGFloat verticalOffset = scrollView.contentInset.top + scrollView.contentInset.bottom;
         CGFloat horizontalOffset =  scrollView.contentInset.left + scrollView.contentInset.right;
-        
+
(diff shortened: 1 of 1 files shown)


WordPress-iOS-Shared/Core/WPNoResultsView.m
```diff
@@ -203,29 +203,40 @@ - (void)showInView:(UIView *)view {
...
-        
+
         CGFloat verticalOffset = scrollView.contentInset.top + scrollView.contentInset.bottom;
         CGFloat horizontalOffset =  scrollView.contentInset.left + scrollView.contentInset.right;
-        
+
+        if ([self.superview isKindOfClass:[UITableView class]]) {
+            UITableView *tableView = (UITableView *)self.superview;
+            CGFloat headerHeight = tableView.tableHeaderView.hidden ? 0 : tableView.tableHeaderView.bounds.size.height;
+            CGFloat footerHeight = tableView.tableFooterView.hidden ? 0 : tableView.tableFooterView.bounds.size.height;
```
aerych: I realize that when the `tableHeaderView` or `tableFooterView` is `nil` the value assigned will resolve to `0`, but the logic just seems a bit wonky, perhaps because there can be nils in the truth test and in the false condition.  Would it make sense to add a nil check as a part of the truth test to at least get the nil out of the false condition?  Something like:
`(tableView.tableHeaderView == nil || tableView.tableHeaderView.hidden)`
Thoughts?

aerych:
Hiya @frosty 
The patch works great!  I had one super nitpicky question.  Up to you to tweak or not.  Regardless, I think we can :shipit:

frosty:
Thanks for the review, @aerych! I made the change as you suggested :)

PR #107, opened 2016-09-02
Fixes font loading.

diegoreymendez:
Even though font loading was seemingly working fine previously, there was a not-very-visible issue with how we were loading them.

We were seeing an issue during Aztec development in which `fontDescriptorWithSymbolicTraits` wasn't able to find the requested fonts and was crashing the app ([due to some internal API issues](http://stackoverflow.com/questions/26622809/why-do-some-ios-font-families-not-support-uifontdescriptortraitbold)).

By switching to loading fonts with `CTFontManagerRegisterFontsForURL`, we now ensure that the loaded fonts live for the duration of the process' life and don't need to be loaded again.  This also means that looking fonts up by trait changes works.

This fixes it.

**More information:**

Apple's documentaton for `CTFontManagerRegisterGraphicsFont` (the method we were previously using) also recommends this switch:

```
Fonts that are backed by files should be registered using CTFontManagerRegisterFontsForURL.
```

Which is what we're pushing in this PR.

**How to test:**

1. Run WPiOS from branch `aztec/integrate-latest`.
2. Enable the Native editor in App Settings.
3. Create a new post and write some content. (the font shown should be Merryweather 16 regular)
4. Select any range of text and make it bold, italic, bold + italic.  Make sure this works and the app doesn't crash.

If you want to see the previous error, modify the Podfile to use Share Pod v0.6.0, and repeat the process - the app WILL crash.


jleandroperez:
Beautiful. Just beautiful.

:shipit:

PR #108, opened 2016-09-29
iOS 10 updates

kurzee:
Updates Xcode 8 recommended project settings and changes the availability of our iOS version checkers to iOS9 and iOS10.

I don't really like the checks, but the iOS 10 check is needed for a bug found in WP-iOS readability implementation.

Note: Dropping the iOS8 check since the pod requires iOS8 or later anyways.

Needs review: @frosty can you take a peak? 🙏


kurzee:
Thanks @frosty! Pushed some changes for the sample project. Can you take one more quick peak?

PR #110, opened 2016-11-07
TableHeader Views: Fix layout issue, cleanup.

kurzee:
- Drop custom font with size usage on `configureTableViewSectionHeader`.
- Removes our older custom WPTableViewSectionHeaderFooterView.

Needs review: @frosty can you take a peak?


frosty:
Hey @kurzee, looks goodI There's one extra reference to `WPTableViewSectionHeaderFooterView` in `WordPressShared.h` – importing `WPTableViewSectionHeaderFooterView.h`. Not sure how everything still compiles with that there, but hey.

After removing that, :shipit:!

WordPress-iOS-Shared.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name         = "WordPress-iOS-Shared"
-  s.version      = "0.6.3"
+  s.version      = "0.6.4"
```
frosty: Actually, thinking about this now – should this be a larger version update (i.e. not just incrementing the final digit)? It's a backwards-incompatible update, as we're removing functionality?

kurzee:
>Actually, thinking about this now – should this be a larger version update (i.e. not just incrementing the final digit)? It's a backwards-incompatible update, as we're removing functionality?

Good catch @frosty! I thought about this for a bit, and I think we should indeed lean towards a larger version bump here.

kurzee:
Thanks @frosty!

PR #113, opened 2017-01-17
Nullability Specifiers

jleandroperez:
### Details:
We need to add nullability specifiers, so that we can drop about a million `force unwraps` in the main project, plus, avoid another million unneeded casts. **Goal:** Fix all of the new compiler warnings we're getting.

**For convenience** i've made it so it's allowed to call few of the **configure** methods with nil parameters. Reasoning behind is: it's safe, (valid ObjC code), and it saves us a lot of unwrapping.

Needs Review: @astralbodies // @frosty 
(Or anyone around!).

Thanks in advance!

### DIFF
--- a/WordPress-iOS-Shared/Core/WPFontManager.h
+++ b/WordPress-iOS-Shared/Core/WPFontManager.h
@@ -1,35 +1,40 @@
 #import <UIKit/UIKit.h>
 
+
+NS_ASSUME_NONNULL_BEGIN
+
 @interface WPFontManager : NSObject
 
 /// Loads the OpenSans font family for the life of the current process.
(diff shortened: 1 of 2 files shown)


frosty:
Thanks for dealing with this, @jleandroperez! 

However... instead of adding all of these `nonnulls`, can we just wrap those regions of code in `NS_ASSUME_NONNULL_BEGIN` and `NS_ASSUME_NONNULL_END`?

jleandroperez:
Ahhh good call @frosty!! ready for another look, thanks!!

frosty:
:shipit: from me too!

jleandroperez:
Thanks gentlemen!

PR #114, opened 2017-01-18
Add dynamic type sizes to WPStyleGuide

nheagy:
Adding dynamic type sizes to `WPStyleGuide` is a first step toward full dynamic type support.

It breaks, like, very few things :P 

Needs review: @aerych

### DIFF
--- a/WordPress-iOS-Shared/Core/WPStyleGuide.h
+++ b/WordPress-iOS-Shared/Core/WPStyleGuide.h
@@ -4,12 +4,6 @@
 @interface WPStyleGuide : NSObject
 
 // Fonts
-+ (UIFont *)largePostTitleFont;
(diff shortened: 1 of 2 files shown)


nheagy:
@aerych a few changes (methods removed). Look ok?

aerych:
Nice :)

I pointed the app's podfile to load the commits in this branch and took the app for a test drive.  Everything looked good there also.  I think :shipit: 

Question tho.  Would it make sense to add a `feature` branch to point to while we're still adding support to the rest of the app, then merge changes to the shared pod (and bump the pod version) once dynamic type is ready to merge in the app?  Just a though, no strong feelings either way.

nheagy:
I decided we'll be fine just bashing it in ;)

PR #123, opened 2018-04-19
Add tracks event for Stock Photos

ctarda:
These are the new events that we have been adding to Tracks while developing Stock Photos for WordPress-iOS.

`WPAnalytics.h` used to be in the main project, but since it is now part of this pod, we need to move these events here.


WordPressShared/Core/Analytics/WPAnalytics.h
```diff
@@ -122,8 +123,13 @@ typedef NS_ENUM(NSUInteger, WPAnalyticsStat) {
     WPAnalyticsStatMediaLibraryEditedItemMetadata,
     WPAnalyticsStatMediaLibraryPreviewedItem,
     WPAnalyticsStatMediaLibrarySharedItemLink,
-    WPAnalyticsStatMediaLibraryAddedPhoto,
-    WPAnalyticsStatMediaLibraryAddedVideo,
+    WPAnalyticsStatMediaLibraryAddedPhotoViaDeviceLibrary,
+    WPAnalyticsStatMediaLibraryAddedPhotoViaOtherApps,
+    WPAnalyticsStatMediaLibraryAddedPhotoViaStockPhotos,
+    WPAnalyticsStatMediaLibraryAddedPhotoViaCamera,
```
nheagy: Did you consider keeping one event for `MediaLibraryAddedPhoto` and adding a parameter for the `via` instead? It'd work like `track(.mediaLibraryAddedPhoto, parameters: ["via": "apps"])`.

I don't know your use case well enough, but it *looks* like it could be a good fit.

astralbodies: Whatever change you make here make sure it matches what's chosen to be done on the Android side too. I can give you a link to a spreadsheet with all of the tracks events listed on both platforms.

etoledom: @nheagy - Yes! It is one event with different parameters.
The confusing part is that we have different enums, but when you check `eventPairForStat:`, the definition is:

```objc
case WPAnalyticsStatMediaLibraryAddedPhotoViaStockPhotos:
    eventName = @"media_library_photo_added";
    eventProperties = @{ @"via" : @"stock_photos" };
```

@astralbodies We are syncing with Android for all this analytics changes 👍

PR #127, opened 2018-04-20
Adding support for NSAttributedString in WPNoResultsView message

etoledom:
This PR adds support for attributes in the WPNoResultsView's message 'label'.
This is used to show a tappable link in the Stock Photos feature in the Main WPiOS app.

These changes where already reviewed and accepted [here](https://github.com/wordpress-mobile/WordPress-iOS/pull/9136), but since we were working in a feature branch, this changes didn't make it to develop on time.

This is a re-implementation of those changes so they are accessible in the feature branch after updating with develop in the main WPiOS project.

### DIFF
--- a/WordPressShared/Core/Views/WPNoResultsView.m
+++ b/WordPressShared/Core/Views/WPNoResultsView.m
@@ -7,8 +7,8 @@
 
 
 @interface WPNoResultsView ()
-@property (nonatomic, strong) UILabel   *titleLabel;
-@property (nonatomic, strong) UILabel   *messageLabel;
+@property (nonatomic, strong) UILabel       *titleLabel;
+@property (nonatomic, strong) UITextView    *messageTextView;
 @end
 
 @implementation WPNoResultsView
@@ -62,7 +62,7 @@
 - (void)commonInit
 {
     [self configureTitleLabel];
-    [self configureMessageLabel];
+    [self configuremessageTextView];
     [self configureButton];
 }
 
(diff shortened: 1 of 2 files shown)


WordPressShared/Core/Views/WPNoResultsView.m
```diff
@@ -7,8 +7,8 @@
 
 
 @interface WPNoResultsView ()
-@property (nonatomic, strong) UILabel   *titleLabel;
-@property (nonatomic, strong) UILabel   *messageLabel;
+@property (nonatomic, strong) UILabel       *titleLabel;
+@property (nonatomic, strong) UITextView    *messageTextView;
```
jleandroperez: OMG @elibud and @SergioEstevao LOOK!!! EDUARDO IS ONE OF MY KIND!!!!!!

WordPressShared/Core/Views/WPNoResultsView.m
```diff
@@ -186,6 +190,20 @@ - (UIImage *)newButtonBackgroundImage {
...
+
+    [mutableAttributedText addAttribute:NSFontAttributeName value:self.messageTextView.font range:fullTextRange];
+    [mutableAttributedText addAttribute:NSForegroundColorAttributeName value:self.messageTextView.textColor range:fullTextRange];
+
+    NSMutableParagraphStyle *paragraphStyle = [NSMutableParagraphStyle new];
+    paragraphStyle.alignment = self.messageTextView.textAlignment;
+    [mutableAttributedText addAttribute:NSParagraphStyleAttributeName value:paragraphStyle range:fullTextRange];
+
+    return mutableAttributedText;
+}
```
jleandroperez: I've got doubts about this:

- The TextView's font and textColor are applied to the full attributedString text range
- Same for the TextAlignment

Bottom line: this method (kinda) defeats the purpose of having an attributedString property, since... no matter what font or color is passed along, it'll get overwritten.

May i ask you what's the use case / requirement that prompted this one?

WordPressShared/Core/Views/WPNoResultsView.m
SergioEstevao: @etoledom ... We will need to have a talk... :)

jleandroperez:
### Update Mark II:
(Suggestions!!)

1. Construct the string with the target Font + Color + Alignment
2. Then linkify all the things with a snippet like this:

```
extension NSMutableAttributedString {

    private var foundationString: NSString {
        return string as NSString
    }

    func linkify(anchorText: String, link: String) {
        var range = foundationString.range(of: anchorText)
        while range.location != NSNotFound {
            addAttribute(.link, value: link, range: range)

            let minimumRange = NSRange(location: range.location + 1, length: foundationString.length - range.location - 1)
            range = foundationString.range(of: anchorText, options: .caseInsensitive, range: minimumRange)
        }
    }
}

let test = NSMutableAttributedString(string: "Automattic Company Testing", attributes: nil)
test.linkify(anchorText: "Company", link: "http://www.a8c.com")
```

--

Squirrel still stands, this can be improved whenever the WPNoResults replacement is in place!

WordPressShared/Core/Views/WPNoResultsView.m
etoledom: 😱😆

PR #128, opened 2018-04-26
Add events for push notification primer

nheagy:
New events for the new [Prime for Push alert](https://github.com/wordpress-mobile/WordPress-iOS/issues/9185)

If it compiles, it should be good 😀


nheagy:
Thanks @jleandroperez!

PR #133, opened 2018-05-22
Get rid of Safe-Expectations.

SergioEstevao:
This PR removes the dependency of Safe Expectation from WordPress-iOS-Shared.

Make sure all the tests are running.

### DIFF
--- a/WordPressShared/Core/Utility/DisplayableImageHelper.m
+++ b/WordPressShared/Core/Utility/DisplayableImageHelper.m
@@ -1,6 +1,5 @@
 #import "DisplayableImageHelper.h"
 #import "NSString+Util.h"
-@import NSObject_SafeExpectations;
 
 static const NSInteger FeaturedImageMinimumWidth = 150;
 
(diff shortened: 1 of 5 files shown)


WordPressShared/Core/Utility/DisplayableImageHelper.m
```diff
@@ -10,6 +9,21 @@
...
+        NSNumber *number = (NSNumber *)obj;
+        result = [number integerValue];
+    } else if ([obj isKindOfClass:NSString.class]) {
+        NSNumberFormatter *numberFormatter = [[NSNumberFormatter alloc] init];
+        numberFormatter.locale = [NSLocale localeWithLocaleIdentifier:@"en_US_POSIX"];
+        NSNumber *number= [numberFormatter numberFromString:(NSString *)obj];
+        result = [number integerValue];
+    }
+    return result;
+}
```
jleandroperez: Nipticky suggestion! How about adding...

```
@interface NSDictionary ()
- (NSInteger)integerForKey:(NSString *)key;
@end
```

That way, if we needed this snippet anywhere else in the entire POD, we could just reuse it.

WordPressShared/Core/Utility/DisplayableImageHelper.m
```diff
@@ -22,15 +36,18 @@ + (NSString *)searchPostAttachmentsForImageToDisplay:(NSDictionary *)attachments
...
-        if ([content containsString:maybeImage]) {
-            imageToDisplay = maybeImage;
-            break;
+        id obj = attachment[AttachmentsDictionaryKeyURL];
+        if ([obj isKindOfClass:NSString.class]) {
+            NSString *maybeImage = (NSString *)obj;
+            if ([content containsString:maybeImage]) {
+                imageToDisplay = maybeImage;
+                break;
+            }
```
jleandroperez: Same comment as above, how about...

```
@interface NSDictionary ()
- (NSString *)safeStringForKey:(NSString  *)key;
@end
```

SergioEstevao:
While I see the point of your comment this will be like reimplementing the
Safe_Expectations methods again, so I want to avoid this.
Hopefully we will move this code to Swift in the future and avoid this
altogether.

I will merge like it is now.

PR #135, opened 2018-05-24
Add new analytics events for Reader Save for Later

frosty:
This PR just adds a couple of analytics events for Reader Save for Later. These are for:

* Saving a post
* Unsaving a post
* Opening a saved post
* Viewing the list of saved posts

**To test:**

* Just check the code looks right, and build the framework :)


frosty:
Thanks!

PR #140, opened 2018-05-30
Brings master current with develop

aerych:
Merges develop back into master. 

@SergioEstevao could I trouble you for a safety check?


aerych:
> it looks develop is on version 1.0.5 so this looks it's just a merge to master of the current version

Exactly that. :) 

Thanks for the extra check @SergioEstevao :)

PR #142, opened 2018-06-14
Add new alerts for notification tracking

nheagy:
These new events will allow us to track push notification interactions with the OS notification alert.


nheagy:
Thanks @aerych!

PR #143, opened 2018-06-29
Adds new keys for support identity and activity log rewind

aerych:
Adds new keys for support identity, and for activity log/rewind

Needs Review: @ScoutHarris would you please?


aerych:
Thank you!

PR #151, opened 2018-09-18
Update to Swift 4.2

ctarda:
@jleandroperez do you want to take a look at this? 

This is the migration to Swift 4.2


ctarda:
I've pushed another update to bump the swift version in `.swift-version`. It took me forever to figure that was the reason why the pod was not listing 🤦‍♂️

PR #152, opened 2018-09-27
Added missing analytics events

yaelirub:
Please refer to: https://github.com/wordpress-mobile/WordPress-iOS/pull/10199


WordPressShared.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressShared"
-  s.version       = "1.1.0"
+  s.version       = "1.1.1-beta.4"
```
jklausa: this is because it took so long for us to review your changes, so it's a 100% on us, but we've already released `  `1.1.1-beta.4`, so this would need to be `.5`. sorry!

yaelirub: 👍 updating

yaelirub:
updated and rebased

PR #158, opened 2018-10-30
Rich Content: Do not filter table tags from rich content.

aerych:
For context see https://github.com/wordpress-mobile/WordPress-iOS/issues/10333

Support for tables in NSAttributedStrings isn't perfect but seems good enough at this point to no longer treat tables as forbidden tags.  This PR removes the regex that would filter table tags from the passed content string.

To Test:
Run the tests, particularly `RichContentFormatterTests` and confirm tests pass.  

I'm not bumping the podspec version in this PR (yet) in case there are other changes we want to include in the next version. 

@nheagy game for a quick review?


aerych:
Thank ya sir!

PR #164, opened 2018-11-12
New Analytics Stat for Properly Placed Pages set parent actions

danielebogo:
This PR introduces 3 new Analytics stat which are used for to track:
• `WPAnalyticsStatPageSetParentViewed` = Set parent page is viewed
• `WPAnalyticsStatPageSetParentSearchAccessed` = Set parent page search is activated
• `WPAnalyticsStatPageSetParentDonePressed` = Set parent page Done button is pressed


WordPressShared.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressShared"
-  s.version       = "1.4.0"
+  s.version       = "1.5.0-beta.1"
```
ScoutHarris: Since this is a minor change, should the version be `1.4.1-beta.1`?

danielebogo: @ScoutHarris  I followed the SemVer approach, changing the minor version instead of the patch version, since this PR adds new events and is not bug fix related.

ScoutHarris: Gotcha. And thanks for the explanation via Slack. I see what you did there. 😄

PR #165, opened 2018-11-13
Text: Removes problematic markup from Gutenberg galleries.

aerych:
In this PR I'm adding some logic to parse away the unordered list markup that surrounds images in Gutenberg authored galleries.  This should remove a list item bullet from appearing next to Gutenberg gallery images in the iOS reader.

To test:
Run the unit test and confirm it passes. 
For the adventurous you might point the iOS app at the commit and rebulid pods. You could also clean your build directory, apply the changes in `RichContentFormatter.swift` to the pod source in the WPiOS project, and then build and run. 

View any post with a Gutenberg gallery. @rachelmcr links to one in https://github.com/wordpress-mobile/WordPress-iOS/issues/10206#issuecomment-436964727 and confirm the bullets do not show and images are correctly sized.

Example:
simulator screen shot - iphone 8 - 2018-11-13 at 17 33 05

@etoledom could I trouble you with this one too?


etoledom:
Here you can see an example post with a gallery and one image with an added caption:
https://etoledomatomicsite01.blog/2018/11/08/gallery-test-v2/

aerych:
Heya @etoledom!  Nice catch on the caption issue.  I should have thought about that.
I've tweaked the regex to allow for it. The consequence is the `figure` tags will also be captured but as far as I can tell they are inconsequential as far as TextKit is concerned. 
Ready for another peek!

aerych:
> it might be a good idea to add an extra test for the figcaption

Excellent idea.  I've added one, and a few comments to clarify the purpose of each assert.  

Thanks @etoledom !!!  When BB is happy I'll merge it up.

PR #176, opened 2019-01-30
Remove FormatterKit.

yaelirub:
As per @stevebaranski suggestion in p77Llu-bxU-p2#comment-12227
Also removed unused function toStringForPageSections.

instead of using the TTTTimeIntervalFormatter for dates that are less than a week ago we use our own relativeStringDate()
This implementation does not include future dates formatting ("from now"). It seems to me like "from now" was not used in the app (looked at scheduled posts)
let me know if I'm wrong and it is used and needed.


WordPressShared/Core/Utility/NSDate+Helpers.swift
```diff
@@ -177,16 +167,6 @@ extension Date {
     public func shortStringWithTime() -> String {
         return DateFormatters.shortDateTime.string(from: self)
     }
-
```
stevebaranski: Nice! "Less code is best code." 😁

yaelirub:
Thanks so much @stevebaranski . Investigating further and will update.

yaelirub:
Hey @stevebaranski . I updated the PR. Please let me know what you think

yaelirub:
@stevebaranski . Thanks for the information. Honestly, I'm not sure anymore. Sounds like FormatterKit might have value still?Especially since DateComponentsFormatter doesn't provide relativity in dates. What do you think?

yaelirub:
Closing this. Opened issue #177

PR #180, opened 2019-02-06
Issues/format video tags

aerych:
Refs https://github.com/wordpress-mobile/WordPress-iOS/issues/10930

This PR adds some new parsing logic to ensure that video tags have the `controls` parameter specified. 

To test: 
Run the new tests and ensure they pass. 
Bonus: Test the WPiOS reader against this update and confirm the video example in the referenced issue plays.

@bummytime could I trouble you for a review?


aerych:
Thank you sir!!

PR #185, opened 2019-02-21
Issues/add stat for tracking excessive load more calls

aerych:
This PR adds a new stat to facilitate some investigation into an [old issue](https://github.com/wordpress-mobile/WordPress-iOS/issues/6819).

@stevebaranski For real this time :)  Could I trouble you for a review?


aerych:
Thanks Steve!
> Please let me know if you need 👀 on the WPiOS side of things

Will do!

PR #190, opened 2019-02-27
Quick Start: cleanup remove dialog events

nheagy:
This PR makes dedicated events for both positive and negative versions of `WPAnalyticsStatQuickStartRemoveDialogButtonTapped`.


nheagy:
Thanks for the quick review @stevebaranski!

PR #202, opened 2019-04-24
Moves WPiOS's Debouncer into WPiOS Shared.

diegoreymendez:
## Description

This PR:

- Moves our `Debouncer` from WPiOS into the shared framework.
- Adds a method to be able to cancel scheduled calls.
- Adds a unit test to test the new `cancel()` method.

The reason for this move is that we need to use the `Debouncer` in https://github.com/wordpress-mobile/WordPressAuthenticator-iOS.  The idea is to use the Debouncer to implement the error delay requested here: https://github.com/wordpress-mobile/WordPress-iOS/issues/10294#issuecomment-486260221

Important: once this is added to the shared library and a new version released, I will submit another PR to remove the Debouncer from WPiOS.

## Testing

Just build and run the unit tests.
Make sure the changes look fine.

### DIFF
--- a/WordPressShared.podspec
+++ b/WordPressShared.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressShared"
-  s.version       = "1.7.4-beta.1"
+  s.version       = "1.7.4-beta.2"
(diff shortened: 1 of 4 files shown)


diegoreymendez:
Thanks!

PR #206, opened 2019-05-29
Removes a stat that's no longer used.

diegoreymendez:
Reference: https://github.com/wordpress-mobile/WordPress-iOS/issues/11809

Changes:
- Stat `wpios_enhanced_site_creation_completed` will no longer be bumped.
- We're going back to bumping stat `wpios_site_created`.

### Testing:

1. Check out WPiOS branch `issue/11809-roll-back-to-bumping-site-created` (which was updated to point to the last commit in this branch).
2. Install the pods.
3. Make sure the App builds correctly.

### DIFF
--- a/WordPressShared/Core/Analytics/WPAnalytics.h
+++ b/WordPressShared/Core/Analytics/WPAnalytics.h
@@ -134,6 +134,7 @@ typedef NS_ENUM(NSUInteger, WPAnalyticsStat) {
     WPAnalyticsStatEnhancedSiteCreationSuccessLoading,
     WPAnalyticsStatEnhancedSiteCreationSuccessPreviewViewed,
     WPAnalyticsStatEnhancedSiteCreationSuccessPreviewLoaded,
+    WPAnalyticsStatEnhancedSiteCreationSuccessPreviewOkButtonTapped,
     WPAnalyticsStatEnhancedSiteCreationErrorShown,
     WPAnalyticsStatGiphyAccessed,
     WPAnalyticsStatGiphySearched,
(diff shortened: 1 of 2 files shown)


WordPressShared/Core/Analytics/WPAnalytics.h
```diff
@@ -134,7 +134,6 @@ typedef NS_ENUM(NSUInteger, WPAnalyticsStat) {
     WPAnalyticsStatEnhancedSiteCreationSuccessLoading,
     WPAnalyticsStatEnhancedSiteCreationSuccessPreviewViewed,
     WPAnalyticsStatEnhancedSiteCreationSuccessPreviewLoaded,
-    WPAnalyticsStatEnhancedSiteCreationCompleted,
```
aerych: Would it make sense to match what @malinajirka is doing with the Android patch and add a `WPAnalyticsStatEnhancedSiteCreationPreviewOKButtonTapped` event?  Not sure if the flows are similar enough to warrant it?

diegoreymendez: Good point, I've added the stat, and updated the WPiOS branch that points to this one.

Please keep in mind the WPiOS branch still doesn't track this, but it's been updated to build - which should be enough for now, and for the purpose of just testing these changes.

diegoreymendez:
@aerych - I've updated this PR with your suggestion, and contrary to what my comment above says, I've also updated the tracking in the WPiOS branch.

Ready for another look!

diegoreymendez:
Thanks @yaelirub and @aerych!

PR #208, opened 2019-06-07
Updates WordPress-iOS-Shared to have better support for Dynamic Type.

diegoreymendez:
This PR is the first step to fixing https://github.com/wordpress-mobile/WordPress-iOS/issues/11874.

Certain styles cause method `fontForTextStyle(_:fontWeight)` to return a font that works well with the initial setup of Dynamic Type, but that fails to update properly when the Dynamic Type size is updated in real time.

This change offers a working alternative for those styles, so that we don't need to manually scale them up.

## Important:

This PR also increases the deployment target to iOS 11, since it's required to use the proposed method.

It's my understanding that this should not be a problem, since we're using the same target in WPiOS.  If there's any other App that needs to be checked please let me know.

## Associated PRs:

WordPressAuthenticator: https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/100
WordPressKit: https://github.com/wordpress-mobile/WordPressKit-iOS/pull/161
WPiOS: https://github.com/wordpress-mobile/WordPress-iOS/pull/11877

## Testing:

Use this PR, and follow the listed instructions to test the changes: https://github.com/wordpress-mobile/WordPress-iOS/pull/11877


diegoreymendez:
Thanks @jkmassel !

PR #209, opened 2019-06-21
Remove table view styling methods

nheagy:
This PR removes the table view styling methods. These will now belong to the apps themselves. I believe that only WPiOS was using these because of the app-specific colors. This will make their use more clear.

See https://github.com/wordpress-mobile/WordPress-iOS/pull/11967


nheagy:
Thanks @danielebogo!

PR #225, opened 2019-11-12
Adds Added Photo Via Media Editor

leandroalonso:
Adds a new Analytics property that refers to Photos that were added from the Media Editor.


yaelirub:
@leandroalonso , are you going to bump the version?

leandroalonso:
@yaelirub Done.

PR #240, opened 2020-01-22
Allow adjustment of Date display strings by Time Zone

bjtitus:
[WPiOS PR #13265](https://github.com/wordpress-mobile/WordPress-iOS/pull/13265#issue-366111364)

Addresses an issue with display strings for posts using the device instead of site time zone: https://github.com/wordpress-mobile/WordPress-iOS/issues/13258

Some additions to `NSDate+Helpers` to adjust the display string for a date based on an optional time zone. The default formatters use the device's time zone but we want to show the site time zone in many cases.

There is some performance penalty for this: about 2x as slow as not manipulating the time zone on the formatters. I think a cache could be added at some point but I'm not sure it's worth doing right now.

Added two tests:
- `testMediumStringRelativeString`: Ensures relative formats are shown. This is not new functionality.
- `testMediumStringTimeZoneAdjust`: Ensures methods are actually adjusting display strings by using the provided time zone.

### DIFF
--- a/WordPressShared.podspec
+++ b/WordPressShared.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressShared"
-  s.version       = "1.8.12-beta.1"
+  s.version       = "1.8.12-beta.2"
   s.summary       = "Shared components used in building the WordPress iOS apps and other library components."
 
   s.description   = <<-DESC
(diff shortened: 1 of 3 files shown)


ScoutHarris:
Hey @pinarol @loremattei .

Can you please include this in the next 14.0 release? It also needs WordPressShared 1.8.12-beta.2 to go with it ([related PR](https://github.com/wordpress-mobile/WordPress-iOS-Shared/pull/240)).

Thanks!!

PR #270, opened 2020-08-03
Makes some properties of AnalyticsEvent public.

diegoreymendez:
Title has it, this PR only makes two properties public for testing purposes.


diegoreymendez:
Thank you!

PR #273, opened 2020-08-10
Bump the pod version to 1.11.0-beta.1 and add comments

mindgraffiti:
This PR bumps the podspec to 1.11.0-beta.1 (instead of `1.10.0-beta.3`) so it can target WPiOS 15.6 and WPAuth 1.23.0

(I'll delete the incorrect 1.10.0-beta.3 tag after this PR is approved and released.)


ScoutHarris (approved):
:shipit:

PR #275, opened 2020-08-31
Load system serif or Noto font at default size for style

frosty:
This PR adds the ability to load a serif font (system-provided on iOS 13+, or Noto on older versions) at the default size for the specified font style. For example, the default size for the `headline` style is 17pt. This is to be used for navigation bar fonts, where we don't want the size to change based on the user's current dynamic type setting.

You can test using the [associated WPiOS PR here](https://github.com/wordpress-mobile/WordPress-iOS/pull/14803).

### DIFF
--- a/WordPressShared.podspec
+++ b/WordPressShared.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressShared"
-  s.version       = "1.11.0"
+  s.version       = "1.12.0-beta.1"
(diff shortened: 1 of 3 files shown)


mindgraffiti:
Don't forget to bump the pod version in the .podspec file!

Gio2018 (commented):
hey @frosty , can you set the new version in the `podspec`? Thanks!

frosty:
Thanks both!

Gio2018 (approved):
LGTM!

PR #280, opened 2020-11-12
Task/add jetpack settings track

yaelirub:
Adding track even for Jetpack settings
Related WPiOS - https://github.com/wordpress-mobile/WordPress-iOS/issues/15275

### DIFF
--- a/WordPressShared.podspec
+++ b/WordPressShared.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name          = "WordPressShared"
-  s.version       = "1.12.0"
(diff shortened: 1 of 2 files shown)


emilylaguna (changes requested):
We recently changed it so you don't have to add new tracks to WordPressShared, even for Objective-C methods. You can just add them to: [WPAnalyticsEvent.swift](https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/WordPress/Classes/Utility/Analytics/WPAnalyticsEvent.swift)

yaelirub:
Implemented in WPAnalyticsEvent.swift as suggested hence closed this one

PR #285, opened 2021-01-17
Add support for m1 machines

jkmassel:
This PR adds M1 machine support by taking care of a few housekeeping things:
- Updating CocoaPods
- Requiring iOS 13

Since I'm doing housekeeping anyway, I also tackled a couple of small things: 
- Updating OHHTTPStubs
- Adjusted methods that are using `FormatterKit` to use `NSRelativeDateFormatter` instead given the fact that `FormatterKit` is deprecated. It slightly changes a couple of strings (as represented in the tests: `just now` will be a localized version of `now`, and `1 day ago` will be replaced with `yesterday`). If these changes aren't ok, we can probably special-case them, but IMHO using system-provided translation is an improvement as it is?


WordPressShared.podspec
```diff
@@ -12,7 +12,7 @@ Pod::Spec.new do |s|
   s.homepage      = "https://github.com/wordpress-mobile/WordPress-iOS-Shared"
   s.license       = "GPLv2"
   s.author        = { "WordPress" => "mobile@automattic.com" }
-  s.platform      = :ios, "11.0"
+  s.platform      = :ios, "13.0"
```
AliSoftware, line 15: :nod:

You said in the PR description that you updated the requirement to iOS 12 but here you're setting it to iOS 13.

I think that's the right move, because `NSRelativeDateFormatter` is iOS13+ so would not work in iOS12. But just wanted to highlight that, given your PR description might set the wrong expectations.

WordPressShared/Core/Utility/NSDate+Helpers.swift
```diff
@@ -127,17 +145,19 @@ extension Date {
...
-        
+
         if let timeZone = timeZone {
             absoluteFormatter.timeZone = timeZone
         }
 
         let components = Calendar.current.dateComponents([.day], from: self, to: Date())
         if let days = components.day, abs(days) < 7 {
-            return relativeFormatter.string(forTimeInterval: timeIntervalSinceNow)
+            return relativeFormatter.localizedString(fromTimeInterval: timeIntervalSinceNow)
```
AliSoftware, line 160: Quick note that another thing this might change compared to the old one is this will say `tomorrow` instead of `in 1 day` (same as your "yesterday" vs "1 day ago", but in other direction, since `abs(days)` works both ways). I think that's ok too (probably unlikely for us to handle a lot of dates in the future I guess… except maybe for showing publishing date of posts that we schedule to be posted in the future) since as you say, better use system-provided localizations for those cases anyway for consistency, but just wanted to highlight this case too.

WordPressShared/Core/Utility/NSDate+Helpers.swift
```diff
@@ -127,17 +145,19 @@ extension Date {
...
-        let relativeFormatter = TTTTimeIntervalFormatter()
+    public func toMediumString(inTimeZone timeZone: TimeZone? = nil) -> String {
+        let relativeFormatter = RelativeDateTimeFormatter()
+        relativeFormatter.dateTimeStyle = .named
+
         let absoluteFormatter = DateFormatters.mediumDate
-        
+
         if let timeZone = timeZone {
             absoluteFormatter.timeZone = timeZone
```
AliSoftware, line 155: ~Even if `RelativeDateTimeFormatter` doesn't have a `timeZone` property, it has a `calendar` property; so shouldn't we set the `relativeFormatter.calendar.timeZone = timeZone` too here?~

nvm, tested it in the Swift REPL and it seems that for RelativeDateTimeFormatter the tz doesn't really matter. Which makes kinda sense after all, since 10 minutes ago is 10 minutes ago in any timezones, since it is formatting a duration / number of seconds between `self` and "now", so those are "two moments in time in the universe" unrelated to timezones…

WordPressSharedTests/NSDateHelperTest.swift
```diff
@@ -33,22 +33,22 @@ class NSDateHelperTest: XCTestCase {
...
-        XCTAssertEqual(date.mediumString(), "just now")
-        XCTAssertEqual(date.addingTimeInterval(-60*5).mediumString(), "5 minutes ago")
-        XCTAssertEqual(date.addingTimeInterval(-60*60*2).mediumString(), "2 hours ago")
-        XCTAssertEqual(date.addingTimeInterval(-60*60*24).mediumString(), "1 day ago")
-        XCTAssertEqual(date.addingTimeInterval(-60*60*24*6).mediumString(), "6 days ago")
+        XCTAssertEqual(date.toMediumString(), "now")
+        XCTAssertEqual(date.addingTimeInterval(-60*5).toMediumString(), "5 minutes ago")
+        XCTAssertEqual(date.addingTimeInterval(-60*60*2).toMediumString(), "2 hours ago")
+        XCTAssertEqual(date.addingTimeInterval(-60*60*24).toMediumString(), "yesterday")
+        XCTAssertEqual(date.addingTimeInterval(-60*60*24*6).toMediumString(), "6 days ago")
```
AliSoftware, line 54: I wonder if we should add future dates in those tests ("tomorrow", "in 2 days", etc). Even if those are rare, I can see stuff like "scheduled publishing dates" being such cases that we could still encounter.

WordPressSharedTests/NSDateHelperTest.swift
```diff
@@ -33,22 +33,22 @@ class NSDateHelperTest: XCTestCase {
...
     }
-    
+
     /// Verifies that  `mediumStringWithTime` takes into account the time zone adjustment
     func testMediumStringTimeZoneAdjust() {
         let date = Date()
         let timeZone = TimeZone(secondsFromGMT: Calendar.current.timeZone.secondsFromGMT() - (60 * 60))
-        XCTAssertEqual(date.mediumString(timeZone: timeZone), "just now")
+        XCTAssertEqual(date.toMediumString(inTimeZone: timeZone), "now")
```
AliSoftware, line 63: Should we stub the current locale in those tests? So that even if the simulator (or the system the tests are run on) is set to use a different locale than US English, the tests will still pass?

(same remark for all the tests in there, not just this one, obv)

AliSoftware (commented):
Overall LGTM, but left a couple of suggestions around tests (stub the system locale, test future dates)

WordPressShared.podspec
jkmassel: Ah yep, good call. I went back and upgraded to iOS 13 last week since now all our projects are there now (they weren't when this PR was first opened) and forgot to update the PR message. Thanks for flagging!

WordPressSharedTests/NSDateHelperTest.swift
jkmassel: Yeah, no objection here – addressed in e9b84e354649c613d6b33067292e5b3bc76607a1.

WordPressSharedTests/NSDateHelperTest.swift
jkmassel: Sure – doesn't hurt!

Addressed in dc42ef1a406a462186ae44cd4b97fc2be8fe4424.

WordPressShared/Core/Utility/NSDate+Helpers.swift
jkmassel: Yeah for sure – added it to the tests in dc42ef1a406a462186ae44cd4b97fc2be8fe4424

WordPressSharedTests/NSDateHelperTest.swift
AliSoftware: That commit fixes the timezone, not the locale, though 🙃

astralbodies:
@jkmassel I bumped Ruby to 2.7.2 as 2.6.4 does not easily compile (well, compile at all) on an arm64 Mac with rbenv. 2.7.2 *just works*. Hopefully that's cool?

astralbodies:
Of course now I'm getting a crash in Ruby... 🤦 

```
➜  WordPress-iOS-Shared git:(add/support-for-m1-machines) bundle exec pod install
Analyzing dependencies
/Users/aarondouglas/src/WordPress-iOS-Shared/vendor/bundle/ruby/2.7.0/gems/ethon-0.12.0/lib/ethon/curls/options.rb:103: [BUG] Segmentation fault at 0x0000000000000002
ruby 2.7.2p137 (2020-10-01 revision 5445e04352) [arm64-darwin20]
```

I think this is the cursed CocoaPods crash people have been talking about. I'm kind of surprised I hadn't run into this with other projects yet. How much of this PR are we assuming is running against arm64 binaries vs. in Rosetta 2? Should Ruby be x86_64?

WordPressSharedTests/NSDateHelperTest.swift
jkmassel: Ummm yes, that is correct. Addressed in 9bb8610bf7f6f4ba6a3fb70e23828337bfe7f601, which converts the project to use a Test Plan, guaranteeing the locale. The time zone change is still useful, because we want to do the tests in UTC.

jkmassel:
> I think this is the cursed CocoaPods crash people have been talking about. I'm kind of surprised I hadn't run into this with other projects yet. How much of this PR are we assuming is running against arm64 binaries vs. in Rosetta 2? Should Ruby be x86_64?

For now, I think it's reasonable to assume folks are only ever using Ruby with `x86_64`, yeah – it works fine and is far more stable, so for the immediate future I don't see a reason not to?

jkmassel:
> Hopefully that's cool?

Yes, it is very cool!

WordPressShared.xcodeproj/project.pbxproj
```diff
@@ -435,6 +436,7 @@
 			children = (
 				827070841ECA4C9500155CBF /* Test Data */,
 				827070851ECA4C9B00155CBF /* Tests */,
+				246BC027260C4C6100C0F7C0 /* TestPlan.xctestplan */,
```
AliSoftware, line 439: Test Plans FTW! 🎉

WordPressSharedTests/NSDateHelperTest.swift
```diff
@@ -23,8 +23,9 @@ class NSDateHelperTest: XCTestCase {
     
     override func setUp() {
         date = dateFormatter.date(from: data.dateString)
+        NSTimeZone.default = TimeZone(secondsFromGMT: 0)!
```
AliSoftware, line 26: 👍

WordPressSharedTests/NSDateHelperTest.swift
```diff
@@ -35,16 +36,27 @@ class NSDateHelperTest: XCTestCase {
     }
 
     /// Verifies that `mediumString` produces relative format strings when less than 7 days have elapsed.
+    /// If this test is failing, check that the Test Plan is still using en-US as its language
```
AliSoftware, line 39: ❤️

AliSoftware (approved):
:shipit:

PR #294, opened 2021-05-18
Fixes the font scaling logic. Fixes a unit test.

diegoreymendez:
Fixes the font scaling logic.  Also fixes a unit test.

For testing instructions please refer to the WordPress-iOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/16520


WordPressShared/Core/Views/WPStyleGuide+DynamicType.swift
```diff
@@ -76,8 +76,7 @@ extension WPStyleGuide {
     @objc public class func fontForTextStyle(_ style: UIFont.TextStyle, maximumPointSize: CGFloat = maxFontSize) -> UIFont {
         if #available(iOS 11, *) {
```
frosty: It looks like WordPressShared targets 11+ now, so we can remove this check.

frosty (approved):
One small note, otherwise :shipit:!

WordPressShared/Core/Views/WPStyleGuide+DynamicType.swift
diegoreymendez: Removed.

PR #297, opened 2021-08-20
Add Buildkite

jkmassel:
Adds Buildkite to this project, and also:
- Switches Ruby back to `2.6.4` for now
- Adds `fastlane` for testing
- Addresses rubocop concerns
- Disables CircleCI publishing


.buildkite/publish-pod.sh
```diff
@@ -0,0 +1,17 @@
...
+install_gems
+
+echo "--- :cocoapods: Publishing Pod to CocoaPods CDN"
+publish_pod $PODSPEC_PATH
+
+echo "--- :cocoapods: Publishing Pod to WP Specs Repo"
+publish_private_pod $PODSPEC_PATH $SPECS_REPO "$SPEC_REPO_PUBLIC_DEPLOY_KEY"
+
+echo "--- :slack: Notifying Slack"
+slack_notify_pod_published $PODSPEC_PATH $SLACK_WEBHOOK
```
mokagio, line 17: I wonder if our plugin could have a `publish_pod` command that takes all the parameters used here as inputs. Just to avoid copy pasting this file in every pod repo. 🤔

The idea is sort of at the edge between useful and not 😅

.buildkite/pipeline.yml
```diff
@@ -0,0 +1,51 @@
...
+    key: "test"
+    command: "build_and_test_pod"
+    env: *common_env
+    plugins: *common_plugins
+    artifact_paths: ".build/logs/*.log"
+
+  #################
+  # Validate Podspec
+  #################
+  - label: "🔬 Validating Podspec"
```
mokagio: Nitpick: 

```suggestion
  - label: "🔬 Validate Podspec"
```

Apart from the command being called "validate", I feel it's more appropriate to use that for the name of the step. Finally, it's what the other steps use 😄

.buildkite/publish-pod.sh
mokagio: Actually... I just noticed that the content for [wpxmlrpc](https://github.com/wordpress-mobile/wpxmlrpc/pull/59) is different. 

Part of me thought: case in point for not extracting this. Another part said: extract and provide and option for public CocoaPods only. Can't make up my mind 🤣

mokagio (approved):
I noticed the publishing via Buildkite failed, but the reason was actually that CircleCI had finished first, so CocoaPods rejected what looked to it like a duplicated publishing request.

I run [a new experiment](https://buildkite.com/wordpress-mobile/wordpress-shared-ios/builds/13) without CircleCI and it succeeded 👍 

Screen Shot 2021-08-23 at 2 53 59 pm


Let me know if you need help getting rid of CircleCI from here. I think this PR does everything that is needed and what's left is only removing the config file?

.buildkite/publish-pod.sh
jkmassel: Yeah, I can see good arguments both ways on this. To me, the end goal is "don't use CocoaPods at all", so I figured "Copy + Paste it is..." 😁

mokagio: > To me, the end goal is "don't use CocoaPods at all", so I figured "Copy + Paste it is..." 😁

💪 I'm with you 👍 Even though I find SPM rather frustrating to use 🙃

PR #300, opened 2021-10-05
Improves the debouncer for ad-hoc callbacks.

diegoreymendez:
Improves the debouncer for ad-hoc callbacks.

Basically we need the debouncer to support the modification of the callback, so that we can specify parameters that are unique to each scheduled call.

## Testing:

To test this, use the WordPress iOS PR.

1. Start the domain registration flow.
2. In the domain suggestions screen, start typing with the device keyboard.

Make sure the whole experience looks fine, and works fine.  The search should trigger after 0.5 seconds since the last keypress.

## Associated PRs:

WordPress iOS PR: https://github.com/wordpress-mobile/WordPress-iOS/pull/17266

### DIFF
--- a/WordPressShared.podspec
+++ b/WordPressShared.podspec
@@ -2,7 +2,7 @@
 
 Pod::Spec.new do |s|
   s.name          = 'WordPressShared'
-  s.version       = '1.16.1'
+  s.version       = '1.16.2-beta.1'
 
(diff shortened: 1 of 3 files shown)


Gio2018 (approved):
LGTM, the unit tests ran successfully. Just remember to do all the bookkeeping and merge it along with the wpios PR.

PR #305, opened 2021-11-29
Merge 1.17.0 into `trunk`

mokagio:
No test required other than verifying the `.podspec` changes and that CI passes.


AliSoftware (approved):
- ✅  Commit list makes sense and diff matches what's in those.
 - ✅  Version bumped in podspec
 
All good :shipit:

PR #319, opened 2022-11-08
Rewrite one test to move away from Specta

crazytonyli:
Mainly because Specta doesn't support SPM. Also, it's [no longer in active development](https://github.com/specta/specta#status).

The test case `WPAnalyticsTests` is also rewritten in Swift, so that we can continue use the Swift version of Quick when  adding Swift Package Manager support later.


WordPressSharedTests/WPAnalyticsTests.swift
```diff
@@ -0,0 +1,133 @@
+import Quick
+
+@testable import WordPressShared
+
+class WPAnalyticsMethodBehaviorExamplesConfiguration: QuickConfiguration {
+    override class func configure(_ configuration: QCKConfiguration!) {
+        sharedExamples("a WPAnalyticsTracker method") { context in
+            guard let context else {
+                XCTFail("Test case needs to provide a context")
+                return
```
mokagio, line 10: Side note: Once can also inline this, like `return XCTFail(...)`.

It's probably not a good idea in this test suite because it's not a pattern it uses but I find it neat, specially under the hood. At first look it looks weird but thinking about it it's obviously okay. `XCTFail` returns `Void` so it's totally legit to "return it" because it's the same as `return Void` or  `return`.

🤓

WordPressSharedTests/WPAnalyticsTests.swift
```diff
@@ -0,0 +1,133 @@
...
+            guard let invocation = context()?["invocation"] else {
+                XCTFail("invocation not found in the provided context")
+                return
+            }
+            guard let theInvocation = invocation as? WPAnalyticsMethodBehaviorInvocation else {
+                XCTFail("invocation not found in the provided context")
+                return
+            }
+
+            it("should not be called if tqracker isn't registered") {
```
mokagio, line 21:
```suggestion
            it("should not be called if tracker isn't registered") {
```

WordPressSharedTests/TestAnalyticsTracker.m
```diff
@@ -38,3 +40,31 @@ - (void)trackString:(NSString *)event withProperties:(NSDictionary *)properties
...
+    id trackerMock2 = OCMStrictClassMock([TestAnalyticsTracker class]);
+    id expectation = [trackerMock expect];
+    id expectation2 = [trackerMock2 expect];
+    invocation(expectation);
+    invocation(expectation2);
+    invocation(trackerMock);
+    invocation(trackerMock2);
+    [trackerMock verify];
+    [trackerMock2 verify];
+}
```
mokagio, line 70: Is the reason these have been defined in the Objective-C layer, although used only in the Swift / Quick code, that they use OCMock?

I noticed that `OCMStrictClassMock` is a macro and we could replace it with its definition like `let trackerMock = OCMockObject.mock(for: TestAnalyticsTracker.self)`. But then things fall apart because OCMock relies on Objective-C's dynamism, but in Swift the type system gets in the way. I was not able to call `verify()` on my Swift `trackerMock` instance.

mokagio (approved):
This was one gnarly test to convert from Specta to Quick. Well done!

Do you plant to do the same for [Expecta](https://github.com/specta/expecta)? As far as I can see, we still have it in the `Podfile` but no longer use it.

WordPressSharedTests/TestAnalyticsTracker.m
crazytonyli: > Is the reason these have been defined in the Objective-C layer, although used only in the Swift / Quick code, that they use OCMock?
Yes. I'd like to keep the original tests, but I'm not familiar with OCMock, not sure if there is a way to move all these code into a Swift file.

crazytonyli:
@mokagio Good catch! I just check Expecta is not used in this project and I've deleted it in https://github.com/wordpress-mobile/WordPress-iOS-Shared/pull/319/commits/bea05fadd067bb0d5aaa15f4174b1e23d3fefd9e

Issue #318, opened 2022-10-27
Remove FormatterKit depencendency


mokagio:
I just opened a PR on WordPress iOS that removes a usage of the deprecated `mediumString(timeZone:)` method, https://github.com/wordpress-mobile/WordPress-iOS/pull/19523.

That method was [deprecated to remove our dependency on FormatterKit](https://github.com/wordpress-mobile/WordPress-iOS-Shared/blob/trunk/WordPressShared/Core/Utility/NSDate%2BHelpers.swift#L121-L136). 

Together with `mediumString(timeZone:)`, we also have [`pageSectionFormatter`](https://github.com/wordpress-mobile/WordPress-iOS-Shared/blob/7099c04108e93d8cbd407f9eef86055d5d5b371d/WordPressShared/Core/Utility/NSDate%2BHelpers.swift#L71-L80). which uses FormatterKit. On top of that, [two](https://github.com/wordpress-mobile/WordPress-iOS-Shared/blob/7099c04108e93d8cbd407f9eef86055d5d5b371d/WordPressShared/Core/Utility/NSDate%2BHelpers.swift#L276-L279) [versions](https://github.com/wordpress-mobile/WordPress-iOS-Shared/blob/7099c04108e93d8cbd407f9eef86055d5d5b371d/WordPressShared/Core/Utility/NSDate%2BHelpers.swift#L211-L224) of `toStringForPageSections` that seem to be unused.

I `git grep`ed for those strings in WordPress and WooCommerce iOS and did not find any match. We should be able to remove them and FormatterKit. This will require a new major version release, so we might take the occasion to review existing TODOs and [other deprecations](https://github.com/wordpress-mobile/WordPress-iOS-Shared/blob/c28fbc12c062df275783c24941de0ef983f5cb56/WordPressShared/Core/Views/WPStyleGuide.h#L67-L79).

mokagio:
Closed by #320.

PR #320, opened 2022-11-08
Remove FormatterKit as a dependency

crazytonyli:
Because FormatterKit doesn't support Swift Package Manager. And the only method that uses it was marked as deprecated. The deprecated method is still kept though, its implementation is swapped to using the recommended API.

---


spencertransier (approved):
The removal looks good to me! 👍 I confirmed that the NSDateHelperTest.testToMediumStringRelativeString test is still passing after this change.

mokagio:
Brilliant! This closes #318, too 🎉

WordPressShared/Core/Utility/NSDate+Helpers.swift
```diff
@@ -120,19 +108,7 @@ extension Date {
...
-            absoluteFormatter.timeZone = timeZone
-        }
-
-        let components = Calendar.current.dateComponents([.day], from: self, to: Date())
-        if let days = components.day, abs(days) < 7 {
-            return relativeFormatter.string(forTimeInterval: timeIntervalSinceNow)
-        } else {
-            return absoluteFormatter.string(from: self)
-        }
+        toMediumString(inTimeZone: timeZone)
```
mokagio, line 111: What do you think about removing this method altogether?

Given this is a `public` API, doing so would be a breaking change and require a new major version, but I think it's okay since we at Automattic are likely to only active consumers of this library, and the upgrade cost for us is very low.

crazytonyli: Sounds reasonable to me. I remember there are a couple other deprecated methods, which can probably go away all together.

PR #321, opened 2022-11-15
Add Swift Package Manager support

crazytonyli:
Unlike similar PRs to other libraries, I didn't split this SPM support change into two PRs, because the complication of "WPAnalyticsTests". I can't get this test to compile in Swift Package Manager, due to we have to use Swift version of Quick and OCMock (which is Objective-C).

And, I ended up deleting this test case. 🥲 

Can't get it to compile is the main reason of course, but also I don't see this test provides much value. I'm not familiar with OCMock, but just by reading the code, I think the verification logic (invoked from OCMock) seems all about "verifying a method is indeed called when a method is called".

- [x] Check if buildkite test collector continues to work

---


.buildkite/pipeline.yml
```diff
@@ -1,19 +1,19 @@
...
 
 # This is the default pipeline – it will build and test the pod
 steps:
-  ################
-  # Build and Test
-  ################
+  ########################
+  # Validate Swift Package
+  ########################
   - label: "🧪 Build and Test iOS"
```
mokagio: Did you decide to keep the label the same to avoid [the GitHub checks dance we run into WordPressUI-iOS](https://github.com/wordpress-mobile/WordPressUI-iOS/pull/120#issuecomment-1313452771)?

.buildkite/pipeline.yml
```diff
@@ -1,19 +1,19 @@
...
     IMAGE_ID: xcode-14
 
 # This is the default pipeline – it will build and test the pod
 steps:
-  ################
-  # Build and Test
-  ################
+  ########################
+  # Validate Swift Package
+  ########################
```
mokagio: FWIW, my two cents are that we can delete these headers.

Sources/WordPressShared/WordPressShared.swift
```diff
@@ -0,0 +1,5 @@
+#if SWIFT_PACKAGE
+
+@_exported import WordPressSharedObjC
+
+#endif /* SWIFT_PACKAGE */
```
mokagio: Nitpick, given there's currently only one `#if`:

```suggestion
#endif
```

Sources/WordPressSharedObjC/Views/WPStyleGuide.m
```diff
@@ -2,8 +2,20 @@
 #import "WPTextFieldTableViewCell.h"
 #import "WPFontManager.h"
 #import "WPDeviceIdentification.h"
-#import "WPShared-Swift.h"
 
+#if SWIFT_PACKAGE
+
+// A workaround to make the Swift extension definied in WordPressSharedView target visible to this file here.
```
mokagio:
```suggestion
// A workaround to make the Swift extension defined in WordPressSharedView target visible to this file here.
```

fastlane/Fastfile
```diff
@@ -5,8 +5,14 @@ default_platform(:ios)
 platform :ios do
   desc 'Builds the project and runs tests'
   lane :test do
+    xcodebuild(
+      scheme: 'WordPressShared',
+      xcargs: '-resolvePackageDependencies'
+    )
```
mokagio, line 11: ~~Interesting... I assume the tests run didn't work without this.~~

~~Have you tried passing the `xcargs` option to `run_tests`?~~

~~The action has parameters such as `disable_package_automatic_updates` and `skip_package_dependencies_resolution` ([both defaulting to `false`](https://docs.fastlane.tools/actions/scan/)) which makes me think it should be able to handle this without a dedicated step.~~

Okay. Silly me, I should have done my reading before commenting 😓 

I can see the error this addresses from the CI builds, e.g. [this one](https://buildkite.com/automattic/wordpress-shared-ios/builds/97#018479fc-7df6-4fab-8f1a-491effdc5525/461-526).

```
[...] Cannot open "Documentation.docc" as a "Swift Package Folder" because it is already open as a "Folder"." 
[...]
[06:35:08]: ▸ ❌  error: Build service could not create build operation:
  internal error:
    could not generate PIF because the workspace has not finished loading or is still waiting for package resolution
```

Seems like a tooling bug to me. Worth adding a note in the code about it?

mokagio: > Seems like a tooling bug to me 

I noticed CI for this project is on Xcode 14.0. I opened #323 to see if Xcode 14.1 solved the alleged tooling issue.

mokagio (approved):

.buildkite/pipeline.yml
crazytonyli: 🙈 I forgot to change it actually... Addressed in https://github.com/wordpress-mobile/WordPress-iOS-Shared/pull/321/commits/7615808b2973db52a5c24c8fbc298fd4eb36d2be

fastlane/Fastfile
crazytonyli: I have seen similar error in Xcode 13 before, my guess is it's caused by some race condition within Xcode between creating the Swift Package's Xcode project and loading the project. But yeah, hopefully this is no longer an issue in Xcode 14.1.

Another thing is, if this issue continues to show up, it would be a good idea to move this additional command into bash-cache's `validate_swift_package` script. Also, it's probably also worth to make the script self contained, rather than relying on the projects to have a `test` lane.

crazytonyli:
If I understand it correctly, Test Analytics of this project is set to only collect test data of the trunk branch. I'll merge this PR and check out the trunk branch to make sure the Test Analytics set up is not broken.

crazytonyli:
> I'll merge this PR and check out the trunk branch to make sure the Test Analytics set up is not broken.

Yep, it's broken. 🤦‍♂️ 

I've opened an issue for it #324, and I'll look into it later.

fastlane/Fastfile
mokagio: > Also, it's probably also worth to make the script self contained, rather than relying on the projects to have a test lane.

Great idea. In particular if we need additional setup.

PR #322, opened 2022-11-15
An experimental tool to check regular expressions at compile time

crazytonyli:
This is a random idea popped into my head today, not sure if it's worth pursuing, but I had fun learning swift-syntax and swift AST. We can move this tool into the bash-cache plugin or some other places, if we find this tool useful. 😸 

The idea is use a special inline comment `// CHECK-REGEX` to check if the following string constant declaration is a valid regex. Kinda like Swift 5.7's Regex constant compile time check, except we don't have to wait for 2 or 3 years (Swift's Regex is only available on iOS 16).

Here is an example:

```swift
// CHECK-REGEX
let pattern = #"\[[^\]]+\]"#
```

One requirement of this check is the string constant needs to be a raw string literal (without string interpolation), otherwise we can't know the regex at compile time.

The implementation is quite simple. Iterate through all the Swift files in given directory, find a declaration that has this special inline comment, extract the regex from the expression, and finally check the regex by creating a `NSRegularExpression` instance with it.

Since this is an experimental tool, I didn't handle any edge cases (i.e. the inline comment is used on a non-string declaration) and the error reporting probably can be improved too (i.e. include file and line number in the error message).

---

- [ ] I have considered if this change warrants release notes and have added them to the appropriate section in the `CHANGELOG.md` if necessary.

### DIFF
--- a/.buildkite/check_regex.sh
+++ b/.buildkite/check_regex.sh
@@ -1,0 +1,9 @@
+#!/bin/bash -eu
+
+cd Tools
+
+echo '--- :swift: Build'
+swift build --product RegexChecker
+
+echo '--- :face_with_monocle: Check Regular Expressions'
+swift run --skip-build RegexChecker
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -40,6 +40,15 @@ steps:
     plugins: *common_plugins
 
   #################
+  # Check Regex
+  #################
+  - label: ":face_with_monocle: Check Regular Expressions"
+    key: "check-regex"
+    command: ".buildkite/check_regex.sh"
+    env: *common_env
+    plugins: *common_plugins
+
+  #################
   # Publish the Podspec (if we're building a tag)
   #################
   - label: "⬆️ Publish Podspec"
@@ -52,4 +61,5 @@ steps:
       - "test"
       - "validate"
       - "lint"
+      - "check-regex"
     if: build.tag != null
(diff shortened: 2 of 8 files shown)


AliSoftware (commented):
That's a very cool idea and experiment! 😍  Love the idea of using Swift to create tooling and of using SwiftSyntax. 🙂 

TBH I'm not sure it's going to be used much in practice tho 😕; after all if we had invalid regexes—as in, doesn't compile—we'd likely have caught it by seeing crashes at runtime at that point, and since it's opt-in, devs will have to remember to use the magic comment, while I'm not sure they will.

So TL;DR: while for a repo like WP-iOS-Shared (which is seldom updated) it can be fine to enable and merge this, I'm not sure it'd be e.g. worth adding a similar tool to all our iOS repos (e.g. WPiOS, WCiOS) or to spend too much time on this, however fun and pretty that tool is?

That being said, I'd love to see some P2 post explaining this (e.g. "Experiment: How to use SwiftSyntax to build a tool to validate code") to share the experience and knowledge and show how easy could be for us to write Swift tooling using SwiftSyntax for other similar means. I think using it can be super powerful, but as SwiftSyntax can be intimidating to anyone who never played with it at first, a P2 post could show how it can be not that scary after all 😄

crazytonyli:
@AliSoftware Agreed. In practice, when adding a new regex, not sure about others but I always go to [some regex tester website](https://regex101.com/) to ensure I have a valid regex for my use case. I sure don't want to test the regex by building and running an iOS app. Good point about P2 by the way, I'll start a draft when I get some time 👍

PR #325, opened 2022-11-24
Delete CocoaLumberjack

crazytonyli:
The main purpose is removing CocoaLumberjack from all libraries(I'll be making similar PRs to WordPressKit and WordPressAuthenticator later), which will make the SPM support slightly easier and potentially solves some logging set up issues we are having on the app side.

A protocol `WordPressLoggingDelegate` (similar to [tracks' `TracksLoggingDelegate`](https://github.com/Automattic/Automattic-Tracks-iOS/blob/d8d1dee9a2cd99a5990bffe26f5fe8f92fd9ac3d/Sources/Model/ObjC/Common/TracksLogging.h#L5-L13)) is introduced in this PR. All the `DDLogXxx` calls are replaced with new `WPSharedLog` calls which uses the `WordPressLoggingDelegate` instance to do the actual logging. On the app side, we need to provide a implementation of this logging delegate protocol (like [`TracksLogger`](https://github.com/wordpress-mobile/WordPress-iOS/blob/6a57f7c848e9393b6eba6036232f973ba5fdb003/WordPress/Classes/Utility/Logging/TracksLogger.swift)).

---


Sources/WordPressShared/Utility/String+RemovingMatches.swift
```diff
@@ -15,7 +15,7 @@ extension String {
         do {
             regex = try NSRegularExpression(pattern: pattern, options: options)
         } catch {
-            DDLogError(("Error parsing regex: \(error)"))
+            WPSharedLogError("Error parsing regex: \(error)")
```
jkmassel, line 18: Next step: we should deprecate this method in favour of one that `throw`s – printing an error to the log then doing the wrong thing is IMHO not great 😅

jkmassel (approved):

PR #327, opened 2022-11-28
Fix cocoapods generate duplicate UUID

crazytonyli:
This PR resolves the "duplicated UUIDs" warning reported by `pod install` when using the `2.0.0-beta.1` podspec.

The Objective-C target has an include directory (which is SPM's convention) which contains symbolic links to the real header files. All those header files are part of the pods' `source_files`, which leads the generated Xcode target contains duplicated header files–one is the symlink file and the other is the real file. Hence, CocoaPods reports the "duplicated UUIDs" warning since the UUID is generated from file name.

The fix is simply excluding the `include` directory, which is meant to be SPM-only anyway, from podspec's `source_files`.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,7 +52,7 @@ _None._
 
 ### Bug Fixes
 
-- Fix an issue where 'pod install' produces a 'duplicate UUID' warning.
+- Fix an issue where 'pod install' produces a 'duplicate UUID' warning. [#327]
 
 ## [2.0.0-beta.1](https://github.com/wordpress-mobile/WordPress-iOS-Shared/releases/tag/2.0.0-beta.1)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,12 @@ _None._
 
 _None._
 
+## [2.0.0-beta.2](https://github.com/wordpress-mobile/WordPress-iOS-Shared/releases/tag/2.0.0-beta.2)
+
+### Bug Fixes
+
+- Fix an issue where 'pod install' produces a 'duplicate UUID' warning.
+
 ## [2.0.0-beta.1](https://github.com/wordpress-mobile/WordPress-iOS-Shared/releases/tag/2.0.0-beta.1)
 
 ### Breaking Changes
--- a/WordPressShared.podspec
+++ b/WordPressShared.podspec
@@ -26,6 +26,7 @@ Pod::Spec.new do |s|
   s.source_files  = ['Sources/WordPressShared/**/*.swift'] \
     + FileList['Sources/WordPressSharedObjC/**/*.{h,m}'].exclude('Sources/WordPressSharedObjC/include')
   s.public_header_files = 'Sources/WordPressSharedObjC/include', 'Sources/WordPressSharedObjC/WordPressShared.h'
+  s.private_header_files = 'Sources/WordPressSharedObjC/Private/*.h'
   s.resource_bundles = {
     WordPressShared: [
       'Sources/WordPressShared/Resources/*.{ttf,otf,json}',
--- a/WordPressShared.podspec
+++ b/WordPressShared.podspec
@@ -1,21 +1,8 @@
 # frozen_string_literal: true
 
-# rubocop:disable Metrics/BlockLength
+require 'rake'
 
-# Helper to lookup ObjC source files
-module Lookup
-  def self.objc_files
-    directory = Pathname.new('Sources/WordPressSharedObjC')
-    include_dir = directory.join('include')
-    public_header_files = include_dir.glob('*.h')
-                                     .map { |link| link.realpath.relative_path_from(Dir.pwd) }
-    source_files = directory.glob('**/*.{h,m}').reject { |file| (file <=> include_dir) == 1 }
-    {
-      source_files: source_files.map(&:to_s),
-      public_header_files: public_header_files.map(&:to_s)
-    }
-  end
-end
+# rubocop:disable Metrics/BlockLength
 
 Pod::Spec.new do |s|
   s.name          = 'WordPressShared'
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -48,6 +48,12 @@ _None._
 
 _None._
 
+## [2.0.0-beta.2](https://github.com/wordpress-mobile/WordPress-iOS-Shared/releases/tag/2.0.0-beta.2)
```
AliSoftware, line 51: We are not supposed to add the version in the title of the header until we do an actual release, because if the PR were to land later than anticipated and after another PR that happened to add, say, a breaking change, or if we happen not to do a release immediately after this PR lands but waited for some others to be added before doing a release, that forecasted version you put in the header would be affected and possibly not be the version number you initially thought it'd be.

So since we can only know what version number to use (in order to comply to SemVer) at the time we actually do the releases and see the whole set of changes done since last release, the practice is supposed to be to keep the title as `Unreleased` when adding items to the CHANGELOG during individual PRs, and only set the title in the PR doing the release.

---

All that being said, I'm not sure if we have an official stance for betas. Maybe @mokagio  can chime in and give his insight? My personal take would be to keep `Unreleased` title until we do a final version, and thus have in the CHANGELOG only section titles that are final versions, not `*beta*` (in other words, not consider a beta tag to be a release in the sense of the CHANGELOG). And even eventually consider ditching the practice to do `*beta*` tags (because they are also subject to the problem from above about not doing the right SemVer bump (because `SemVerBump(Change1) + SemVerBump(Change2) ≠ SemVerBump(Change1 + Change2)`) and instead have clients point to a commitish (sha1) during development, and only do (non-beta) lib releases (and update the `Unreleased` title) during code freeze. Just like we've been doing for a long while in the Android side. At least I thought that was the plan, just not sure if it has been officialized yet.

CHANGELOG.md
```diff
@@ -48,6 +48,12 @@ _None._
...
+
+### Bug Fixes
+
+- Fix an issue where 'pod install' produces a 'duplicate UUID' warning. [#327]
+
 ## [2.0.0-beta.1](https://github.com/wordpress-mobile/WordPress-iOS-Shared/releases/tag/2.0.0-beta.1)
```
mokagio, line 57: Have you considered merging these in `2.0.0`? Do we need to have beta level granularity or can we get by with only release notes for stable releases?

WordPressShared.podspec
```diff
@@ -21,7 +23,8 @@ Pod::Spec.new do |s|
   s.swift_version = '5.0'
 
   s.source        = { git: 'https://github.com/wordpress-mobile/WordPress-iOS-Shared.git', tag: s.version.to_s }
-  s.source_files  = 'Sources/WordPressShared/**/*.swift', 'Sources/WordPressSharedObjC/**/*.{h,m}'
+  s.source_files  = ['Sources/WordPressShared/**/*.swift'] \
+    + FileList['Sources/WordPressSharedObjC/**/*.{h,m}'].exclude('Sources/WordPressSharedObjC/include')
```
mokagio, line 27: TIL about Rake's [`FileList`](https://docs.ruby-lang.org/en/2.1.0/Rake/FileList.html) 🤓 

Do we need the `\` to go to new line?

```suggestion
  s.source_files  = ['Sources/WordPressShared/**/*.swift']
    + FileList['Sources/WordPressSharedObjC/**/*.{h,m}'].exclude('Sources/WordPressSharedObjC/include')
```

AliSoftware (commented):
LGTM but left a (not directly related) question about CHANGLELOG

CHANGELOG.md
AliSoftware: Lol I should have refreshed the page before submitting my prose, I see @mokagio ended up asking a similar (but more concise 😅) question [here](https://github.com/wordpress-mobile/WordPress-iOS-Shared/pull/327#discussion_r1034624508)… great minds 😄

mokagio: > My personal take would be to keep Unreleased title until we do a final version

I'd prefer this as well.

mokagio (approved):
I don't consider the `CHANGELOG.md` formatting a blocker for merging this PR. 

However, I disabled auto-merge so @crazytonyli it's up to you where to update the format here, or merge as is then use a dedicated PR as the discussion forum for the format to adopt going forward.

CHANGELOG.md
crazytonyli: > We are not supposed to add the version in the title of the header until we do an actual release

I should've mentioned it in the PR description. I took a shortcut here. Instead of waiting for this PR to be merged and creating a separate PR to bump the version and update changelog (like I did for [2.0.0-beta.1](https://github.com/wordpress-mobile/WordPress-iOS-Shared/pull/326)), I did them all in one PR. So that I can release a new beta and unblock https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/708 and https://github.com/wordpress-mobile/WordPressKit-iOS/pull/550 straight after this PR is merged.

I agree with your point about a potentially conflict with other PRs. But in this particular instance, this repo doesn't get PRs very often, which is the major reason I thought it's okay to include the "release changes" in this PR.

In regards to including betas in CHANGELOG.md or not, my thinking is versions in CHANGELOG.md should match the ones in [the GitHub Releases page](https://github.com/wordpress-mobile/WordPress-iOS-Shared/releases). If it's a standard practice to do beta releases, then we need to include betas in the CHANGELOG.md. But I don't have a strong opinion on whether we should do beta or not 😄

WordPressShared.podspec
crazytonyli: Yep, otherwise it's a syntax error. With a `\`, line 26 and 27 are one expression—contacting two arrays and assigning them to a variable. Without it, they are two separate expressions and the second one line 27 is an illegal expression.

crazytonyli:
I'll merge this PR for now, but I can create another PR to remove the beta releases from CHANGELOG.md if needed. 😃

PR #328, opened 2022-12-05
Fix an occasional crash caused by performSelector:withObject:

crazytonyli:
Not sure why this method would cause a crash on `objc_retain`. Also this crash only happens on CI, not on my Mac.

Two out of five jobs on this repo's trunk branch failed because of this issue. Whereas in a WordPressKit PR, this crash happens constantly.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,7 +46,7 @@ _None._
 
 ### Internal Changes
 
-- Fix an occasional crash caused by `performSelector:withObject:`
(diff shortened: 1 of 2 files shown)


mokagio (approved):
I'd be great to dig into this, but also feels very close to the underlying Objective-C runtime. Better spend our time chasing issues in our higher level domain code—there's plenty of them anyway.

Thank you for the fix and documentation, both here and in the other codebase(s) where you applied it.

PR #329, opened 2022-12-09
Move all public headers into the include directory

crazytonyli:
The include directory used to have [symbolic links to the public headers](https://github.com/wordpress-mobile/WordPress-iOS-Shared/tree/2.0.0-beta.2/Sources/WordPressSharedObjC/include), which isn't really necessary and made [the podspec file](https://github.com/wordpress-mobile/WordPress-iOS-Shared/blob/2.0.0-beta.2/WordPressShared.podspec#L27) unnecessarily complicated. This PR deletes all the symbolic links in the include directory and moves all the public header into it.

---


mokagio (approved):
> The include directory used to have [symbolic links to the public headers](https://github.com/wordpress-mobile/WordPress-iOS-Shared/tree/2.0.0-beta.2/Sources/WordPressSharedObjC/include?rgh-link-date=2022-12-09T01%3A18%3A12Z)

😱

crazytonyli:
@mokagio Yeah, that would be my reaction too 🥲. I'm not sure why I went with using symlinks at the first place. Maybe that's the approach I used a long time ago and stuck in my head as "the way" to adopt SPM in Objective-C codebase.

PR #331, opened 2023-01-12
Fix resource bundle for WordPressShared when installed from CocoaPods

dvdchr:
Refs https://github.com/wordpress-mobile/WordPress-iOS/issues/19877, p5T066-3ND-p2#comment-14170, p5T066-3ND-p2#comment-14172

This fixes an issue where the WordPressShared no longer loads the Noto font family correctly, resulting in the app content (editor, comment threads) being displayed with the default font from HTML.

In `1.18.0`, `WPFontManager` loads the bundle was loaded through this method:

https://github.com/wordpress-mobile/WordPress-iOS-Shared/blob/3ed6530c1cb7a56c632afad808b279074692d0a3/WordPressShared/Core/Utility/WPFontManager.m#L113

However, the method has since been [moved to the Swift part](https://github.com/wordpress-mobile/WordPress-iOS-Shared/blob/trunk/Sources/WordPressShared/Utility/NSBundle%2BWordPressShared.swift) in `WordPressShared`, and is inaccessible from `WPFontManager` because the font manager is inside `WordPressSharedObjC` which is depended upon by `WordPressShared`, but not the other way around. CMIIW.

When debugging the app, the fonts are located in `WordPress.app/WordPressShared.bundle` since the `WordPressShared` is still included through CocoaPods. Following the solution from `NSBundle+WordPressShared.swift`, this attempts to add the `WordPressShared.bundle` to the bundle path when loading the font.

## To Test

Refer to https://github.com/wordpress-mobile/WordPress-iOS/pull/19898.

---

### DIFF
--- a/Sources/WordPressSharedObjC/Utility/WPFontManager.m
+++ b/Sources/WordPressSharedObjC/Utility/WPFontManager.m
@@ -108,12 +108,7 @@ static NSString* const NotoRegularFileName = @"NotoSerif-Regular";
 
 + (void)loadFontResourceNamed:(NSString *)name withExtension:(NSString *)extension
 {
-#if SWIFT_PACKAGE
-    NSBundle *bundle = SWIFTPM_MODULE_BUNDLE;
-#else
-    NSBundle *bundle = [NSBundle bundleForClass:[self class]];
-#endif
-    NSURL *url = [bundle URLForResource:name withExtension:extension];
+    NSURL *url = [[self resourceBundle] URLForResource:name withExtension:extension];
 
     CFErrorRef error;
     if (!CTFontManagerRegisterFontsForURL((CFURLRef)url, kCTFontManagerScopeProcess, &error)) {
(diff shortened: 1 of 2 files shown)


crazytonyli:
Ops, my bad, not sure why I [replaced the original `wordPressSharedBundle` call with `bundleForClass:`](https://github.com/wordpress-mobile/WordPress-iOS-Shared/pull/321/files#diff-e7a1147c8e65869458abbe27d03da80069563c40b99ac7733747386001117d82R114)... Thanks for the fix!

Sources/WordPressSharedObjC/Utility/WPFontManager.m
```diff
@@ -125,4 +120,20 @@ + (void)loadFontResourceNamed:(NSString *)name withExtension:(NSString *)extensi
...
 }
 
++ (NSBundle *)resourceBundle {
+#if SWIFT_PACKAGE
+    return SWIFTPM_MODULE_BUNDLE;
+#else
+    // When installed via CocoaPods, the resources will be bundled under `WordPressShared.bundle`.
+    // This follows the implementation from `NSBundle+WordPressShared`.
+    NSBundle *defaultBundle = [NSBundle bundleForClass:[self class]];
+    NSURL *sharedBundleURL = [defaultBundle.bundleURL URLByAppendingPathComponent:@"WordPressShared.bundle"];
```
crazytonyli: The Swift version [uses `resourceURL`](https://github.com/wordpress-mobile/WordPress-iOS-Shared/blob/3defd5e1a0b96e3cce0ee83163746563d48c6490/Sources/WordPressShared/Utility/NSBundle%2BWordPressShared.swift#L19), instead of `bundleURL`, should we keep them consistent?

crazytonyli (changes requested):

Sources/WordPressSharedObjC/Utility/WPFontManager.m
dvdchr: Good catch. This is addressed in https://github.com/wordpress-mobile/WordPress-iOS-Shared/pull/331/commits/623fc90c6ae9c8a4b1dd96d4379e0a18cb4cb415 🙂

crazytonyli (approved):

PR #335, opened 2023-01-28
Add a `WordPressLoggingDelegate` implementation that logs to console

mokagio:
When working on the WordPress Authenticator  demo app for Google SignIn testing, I got a message about `WordPressLoggingDelegate` not being implemented. I put together this simpler implementation to `print`s to the console to satisfy the requirement. I think it might be useful during development or debugging and worth shipping together with the `protocol`.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,7 +38,7 @@ _None._
 
 ### New Features
 
-_None._
+- Add `ConsoleLogger`, a `WordPressLoggingDelegate` implementation that can be used during development and debugging [#335]
 
 ### Bug Fixes
 
--- a/Sources/WordPressShared/Utility/ConsoleLogger.swift
+++ b/Sources/WordPressShared/Utility/ConsoleLogger.swift
@@ -1,0 +1,24 @@
+/// A `WordPressLoggingDelegate` implementation that logs to the Xcode console via `print`.
+/// Useful for development or debugging. Not recommended in release builds.
+public class ConsoleLogger: NSObject, WordPressLoggingDelegate {
+
+    public func logError(_ str: String) {
+        print("❌ – Error: \(str)")
+    }
+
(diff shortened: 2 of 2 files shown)


crazytonyli (approved):
Love the emojis :shipit: 

BTW, I feel like `NSLog` would give us more (sometimes quite important) info in the logs, like timestamp, thread id, etc. My personal preference is `NSLog` over `print`, but I don't really have a strong opinion on this.

mokagio:
Thank you for the `NSLog` suggestion @crazytonyli. It does provide more info indeed, at no extra cost:

```
2023-01-30 14:37:07.251471+1100 AuthenticatorDemo[47512:28088466] 📃 – Verbose: Received Social Login Oauth response.
```

vs 

```
📃 – Verbose: Received Social Login Oauth response.
```

Implemented in [wordpress-mobile/WordPress-iOS-Shared@`cd428c3` (#335)](https://github.com/wordpress-mobile/WordPress-iOS-Shared/pull/335/commits/cd428c3fcda83b36977eaffc795767ed358c43d6)

mokagio:
By the way @crazytonyli, I wanted to run the tests locally to see if my changes compiled but I got the following error from Fastlane:

```
xcodebuild: error: The workspace named "WordPressShared" does not contain a scheme named "WordPressShared". The "-list" option can be used to find the names of the schemes in the workspace.
```

I wonder if you get a similar behavior on your end, but I suspect there must be something wrong on my side as CI wouldn't be green otherwise.

crazytonyli:
@mokagio It's likely caused by the left over xcworkspace or xcodeproj file, which were deleted after adding SPM support. If those files are still there, the `xcodebuild` command would pick them up as the project to build, instead of the `Package.swift`.

mokagio:
Spot on. I should have realized 🤦‍♂️ Thanks @crazytonyli !

PR #344, opened 2024-01-19
Add SwiftLint via SPM

mokagio:
Adds SwiftLint as a [Swift Package Plugin](https://developer.apple.com/videos/play/wwdc2022/110359/) and adds it to CI using the dedicated step on Docker with build annotation that we are settling in as our default, see paaHJt-4l3-p2#comment-8443 

**One important caveat.** There is what to me looks like a SwiftLint issue that makes the integration fail to build when using a remote configuration. We ought to do more investigation on this front, but, in the interest of getting this over the line and iterated upon, I decided to copy paste our remote configuration here.

Notice that before we can merge this PR we'll need to merge https://github.com/wordpress-mobile/WordPress-iOS-Shared/pull/347 into it.

I decided to split the PRs so that the one setting up the tooling didn't have any noise from the fixes and also worked as a demo for how the integration picks up violations in CI.

---

- [x] I have considered if this change warrants release notes and have added them to the appropriate section in the `CHANGELOG.md` if necessary. – N.A. this is an internal change


.buildkite/commands/swiftlint.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash -u
+
+echo "--- :swift: Running SwiftLint"
+
+set +e
+SWIFTLINT_OUTPUT=$(swiftlint lint --quiet "$@" --reporter relative-path)
```
crazytonyli, line 6: There is a `markdown` reporter in swiftlint. Maybe we can try that and use the output as buildkite annotations?

mokagio: @crazytonyli I opened https://github.com/wordpress-mobile/WordPress-iOS-Shared/pull/348 to test this. It's failing for some reason I haven't yet figured out.

In the meantime, I looked at the output locally and found a few things that make me feel like we should stick to the current approach.

Sample:

file | line | severity | reason | rule_id
--- | --- | --- | --- | ---
/Users/gio/Developer/a8c/WordPress-iOS-Shared/Tests/WordPressSharedTests/NSDateHelperTest.swift | 10 | :stop\_sign: | Trailing Whitespace: Lines should not have trailing whitespace | trailing_whitespace
/Users/gio/Developer/a8c/WordPress-iOS-Shared/Tests/WordPressSharedTests/NSDateHelperTest.swift | 15 | :stop\_sign: | Trailing Whitespace: Lines should not have trailing whitespace | trailing_whitespace
/Users/gio/Developer/a8c/WordPress-iOS-Shared/Tests/WordPressSharedTests/NSDateHelperTest.swift | 23 | :stop\_sign: | Trailing Whitespace: Lines should not have trailing whitespace | trailing_whitespace
/Users/gio/Developer/a8c/WordPress-iOS-Shared/Tests/WordPressSharedTests/NSDateHelperTest.swift | 64 | :stop\_sign: | Trailing Whitespace: Lines should not have trailing whitespace | trailing_whitespace
/Users/gio/Developer/a8c/WordPress-iOS-Shared/Tests/WordPressSharedTests/NSDateHelperTest.swift | 69 | :stop\_sign: | Trailing Whitespace: Lines should not have trailing whitespace | trailing_whitespace
/Users/gio/Developer/a8c/WordPress-iOS-Shared/Tests/WordPressSharedTests/NSDateHelperTest.swift | 74 | :stop\_sign: | Trailing Whitespace: Lines should not have trailing whitespace | trailing_whitespace


The paths are absolute and there is no code fencing around them. Also, the output uses `:stop\_sign:` and `:warning:` as emojis to represent errors and warnings respectively. I love emojis, but I wonder if rows of them might make the output too heavy.

All in all, I think what @iangmaia has configured here with some simple formatting is a better annotation.

Of course, we could add formatting code to the `markdown` output, but I feel it's not worth the additional effort.

Something else I thought of was making a custom reporter, e.g `--reporter buildkite` but 1) that would be even more effort 😅 2) I think the only way to use it would be for it to be merged into SwiftLint, which would take time.

iangmaia: I've also looked into the `markdown` output, which doesn't look very good imo. I had a brief discussion about it and other options [here](https://github.com/Automattic/pocket-casts-ios/pull/1326#discussion_r1457609311).

crazytonyli (approved):

PR #351, opened 2024-02-23
RFC: Load SwiftLint version in the Package Description

iangmaia:
This PR is a simple experiment on loading the right SwiftLint version defined in `.swiftlint.yml` in the Package Description.
I have refrained from using a proper YAML parser to keep it simple, so I just load the file and parse it as a string.

I don't have many ideas on sharing this code to use in other repos, but perhaps it also isn't a big deal to replicate it as needed.

---

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,5 +1,6 @@
 // swift-tools-version:5.5
 
+import Foundation
 import PackageDescription
 
 let package = Package(
@@ -14,7 +15,7 @@ let package = Package(
         // See https://github.com/erikdoe/ocmock/issues/500#issuecomment-1002700625
         .package(url: "https://github.com/erikdoe/ocmock", revision: "afd2c6924e8a36cb872bc475248b978f743c6050"),
         .package(url: "https://github.com/Quick/Quick", from: "6.0.0"),
-        .package(url: "https://github.com/realm/SwiftLint", from: "0.54.0")
+        .package(url: "https://github.com/realm/SwiftLint", .exactItem(loadSwiftLintVersion()))
     ],
     targets: [
         .target(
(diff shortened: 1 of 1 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true" data-kind="Warning">
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false"><code>Package.swift</code> was changed without updating its corresponding <code>Package.resolved</code>. Please resolve the Swift packages in Xcode.</td>
    </tr>
  </tbody>
</table>

<p align="right" data-meta="generated_by_pr-check">
  Generated by :no_entry_sign: <a href="https://danger.systems/">Danger</a>
</p>

Package.swift
```diff
@@ -60,3 +61,25 @@ let package = Package(
...
+
+    // assumes the format `swiftlint_version: <version>`
+    guard let version = Version(versionLine.components(separatedBy: ":")
+                                    .last?
+                                    .trimmingCharacters(in: .whitespaces) ?? "") else {
+        fatalError("Failed to extract SwiftLint version.")
+    }
+
+    return version
+}
```
mokagio: Love it. 

It's a bummer we'll have to add it to every `Package.swift`. But, any centralization attempt would be vain because we'd need a different dependency manager to fetch it ♻️ .

One question, I'm not sure what the rule behind the indentation is.

I run it in Xcode and got:

```suggestion
func loadSwiftLintVersion() -> Version {
    guard let yamlString = try? String(contentsOf: URL(fileURLWithPath: #file)
        .deletingLastPathComponent()
        .appendingPathComponent(".swiftlint.yml")) else {
        fatalError("Failed to read YAML file.")
    }

    guard let versionLine = yamlString.components(separatedBy: .newlines)
        .first(where: { $0.contains("swiftlint_version") }) else {
        fatalError("SwiftLint version not found in YAML file.")
    }

    // Assumes the format `swiftlint_version: <version>`
    guard let version = Version(versionLine.components(separatedBy: ":")
        .last?
        .trimmingCharacters(in: .whitespaces) ?? "") else {
        fatalError("Failed to extract SwiftLint version.")
    }

    return version
}
```

I point this out only because if we replicate this code in other projects, I would be good to have it consistent.

mokagio (approved):
Looking good. I tried changing the version locally and it did indeed fetch the new one. 👍

Package.swift
iangmaia: > But, any centralization attempt would be vain because we'd need a different dependency manager to fetch it ♻️ .

I was wondering if another package acting as a wrapper plugin around SwiftLint and adding this extra functionality would work. Perhaps something to try out in the future.

